PR #8, opened 2024-05-14
WidgetSet so allow selecting the appropriate widget with buttons

ericthelin:
make buttons and widget switching work


tecnovel (changes requested):
i will commit almost the same version but with global code design for local vars-> m_... instead of only _...

tecnovel (approved):

PR #9, opened 2024-05-15
Caching weather and Timezones

ericthelin:
This implements a simple caching system for the weather widget so that the weather is only updated on an interval and is kept in the widget for redisplay if you quickly cycle widgets. It also fixes time zone implementation for both clock and weather


tecnovel (changes requested):
1. i notice there is a new small delay when the code boots up between connecting to the wifi and the clock shows up. like it shows first only the shadows and then the clock. Is this new?
2. if you switch to the weather widget later (not the first time) it refreshes the text, temperature and 3days forecast twice without any reason to refresh (no text changes)

if you could look at these two things would be great! give me feedback if you don't see it or if you have a question

tecnovel (approved):

PR #10, opened 2024-05-16
Porting over of the Stock Widget to the new code.

OoniPizza:
…ckScreenWidget.

Trello Card: https://trello.com/c/SHARECODE


Info-Orbs/lib/config/config.h.template
```diff
@@ -82,12 +82,12 @@
...
 
-#define TIME_ZONE_OFFSET 2
+#define TIME_ZONE_OFFSET -4
 #define FORMAT_24_HOUR true
 
-#define WEATHER_LOCAION "Aberdeen,ID";
+#define WEATHER_LOCAION "47542";
 #define WEATHER_UNITS "us";
-#define WEATHER_API_KEY "XW2RDGD6XK432AF25BNK2A3C7"
+#define WEATHER_API_KEY "L8P9DU7SYN3DZ7QEBWY7YB8FR"
```
ericthelin, line 90: The exiting api key is Brett's that he wanted to stay hard coded.

Info-Orbs/src/main.cpp
```diff
@@ -7,7 +7,7 @@
 #include <Button.h>
 #include <globalTime.h>
 #include <config.h>
-#include <widgets/screenWidgets/stockScreenWidget.h>
+#include <widgets/stockScreenWidget.h>
```
ericthelin, line 10: I think the widget should simply be `stockWidget` now as the Screen part is meaningless.

Info-Orbs/src/main.cpp
```diff
@@ -40,7 +40,7 @@ bool tft_output(int16_t x, int16_t y, uint16_t w, uint16_t h, uint16_t *bitmap)
 
 ScreenManager* sm;
 WidgetSet* widgetSet;
-StockScreenWidget *stockScreenWidget;
+//StockScreenWidget* stockScreenWidget;
```
ericthelin, line 43: I think this can be fully removed instead of just commented out

Info-Orbs/include/widgets/stockScreenWidget.h
```diff
@@ -0,0 +1,34 @@
...
+
+    // Stock data variables
+    String m_TrackingStock;
+    String currentPrice;
+    String change;
+    String percentChange;
+    String highPrice;
+    String lowPrice;
+    String openPrice;
+    String previousClosePrice;
```
ericthelin, line 31: I believe the pattern is to prefix all private member variables with `m_`. I know you were copying what was in the weatherWidget but it is wrong and will be changed in the near future.

Info-Orbs/src/widgets/stockScreenWidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    String httpRequestAddress = "https://finnhub.io/api/v1/quote?symbol=" + m_TrackingStock + "&token=c1tjb52ad3ia4h4uee9g";
+
+    HTTPClient http;
+    http.begin(httpRequestAddress);
+    int httpCode = http.GET();
+    
+    if (httpCode > 0) { // Check for the returning code
+        String payload = http.getString();
+        StaticJsonDocument<1024> doc;
```
ericthelin, line 41: is there an advantage of StaticJsonDocument over simply `JsonDocument doc`?

Info-Orbs/src/widgets/stockScreenWidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+        DeserializationError error = deserializeJson(doc, payload);
```
ericthelin, line 42: very nice! I will be updating the weatherWidget shortly to use this.

Info-Orbs/src/widgets/stockScreenWidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+        displayStock(currentStockIndex, TFT_WHITE, TFT_BLACK);
+        m_stockDelayPrev = millis();
+        currentStockIndex = (currentStockIndex + 1) % stockTickers.size();
+    }
+}
+
+void StockScreenWidget::update() {
+}
+
+void StockScreenWidget::getStockData() {
```
ericthelin, line 29: what if you put this all into the update() method and had everything store to an array so that you keep the data for the next time it is displayed? This is what the weatherWidget is now doing. And it allows cycling through widgets without forcing new data every time if the data is still current.

Info-Orbs/src/widgets/stockScreenWidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+    // Get initial stock then every m_stockDelay milliseconds
+    if (m_stockDelayPrev == 0 || (millis() - m_stockDelayPrev) >= m_stockDelay) {
+        getStockData();
+        displayStock(currentStockIndex, TFT_WHITE, TFT_BLACK);
+        m_stockDelayPrev = millis();
+        currentStockIndex = (currentStockIndex + 1) % stockTickers.size();
+    }
+}
+
+void StockScreenWidget::update() {
```
ericthelin, line 26: This will need to be updated to use the new `force` bool that was just merged. At least to implement the new method signature.

Info-Orbs/src/widgets/weatherWidget.cpp
ericthelin, line 1: These changes should be done in a separate PR

Info-Orbs/lib/config/config.h.template
OoniPizza: I agree... I added mine before I saw that post.

Info-Orbs/src/main.cpp
OoniPizza: I was thinking about that as well.... You may have had it commented out with that names and I just ran with it. I agree removing "screen" from the name.

Info-Orbs/src/main.cpp
OoniPizza: I agree

Info-Orbs/include/widgets/stockScreenWidget.h
OoniPizza: Good to know... I have no problem with this.

Info-Orbs/src/widgets/stockScreenWidget.cpp
OoniPizza: I am not sure about this one. I think it was something I found via Google. I would have to do some research to understand advantages.

Info-Orbs/src/widgets/stockScreenWidget.cpp
OoniPizza: That sounds good. A little more advance for me at the moment.

Info-Orbs/src/widgets/stockScreenWidget.cpp
OoniPizza: I saw you all were chatting about this. I am fine with whatever you all decide. I would just have to see how to do it.

Info-Orbs/src/widgets/weatherWidget.cpp
OoniPizza: This is the first time doing this. I agree I could have done these changes in a separate request. I am still learning.

Info-Orbs/src/widgets/stockScreenWidget.cpp
ericthelin: look at the current WeatherWidget it is using `JsonDocument` and it seems to work. I believe you are statically reserving space where the other is dynamically allocating the space. dynamic is simpler.

Info-Orbs/src/widgets/stockScreenWidget.cpp
ericthelin: look at the new version of Widget and either WeatherWidget or ClockWidget.

Info-Orbs/src/widgets/weatherWidget.cpp
ericthelin: no problem. You are doing great!

tecnovel (approved):
looks good to me, but implement as next step the caching of you data, so the stocks can be show directly if you switch to the widget (look at the weather example, save all in a var and detect if there has been changes) and also implement the force logic correctly. also put the getStockData into the update function so we can seperate update and drawing. good first steps! like your effort, if you have any questions just contact me or @ekt242 on discord

PR #11, opened 2024-05-17
Use proper private notation for weather properties and minor fixes

ericthelin:
The minor fixes include handling network and parse errors on weather data.


tecnovel (approved):
looks good

PR #21, opened 2024-05-30
Stock widget update

ericthelin:
https://trello.com/c/VOVaqRYI
https://trello.com/c/ulgNeGBt
https://trello.com/c/xjwMScgJ

## Changes
This is a major update to the Stocks Widget to add proper update frequency and data handling. It also updates data from all widgets at startup with a simple loading screen. Which in turn updates the timezone. I haven't removed the timezone from config yet but it should be safe to do so now.


ericthelin:
wrong branch again...

PR #22, opened 2024-05-30
refactor stock widget to cache data and handle updates better

ericthelin:
https://trello.com/c/VOVaqRYI
https://trello.com/c/ulgNeGBt
https://trello.com/c/xjwMScgJ

## Changes
This is a major update to the Stocks Widget to add proper update frequency and data handling. It also updates data from all widgets at startup with a simple loading screen. Which in turn updates the timezone. I haven't removed the timezone from config yet but it should be safe to do so now.

### DIFF
--- a/Info-Orbs/include/widgets/stockWidget.h
+++ b/Info-Orbs/include/widgets/stockWidget.h
@@ -23,7 +23,7 @@ class StockWidget : public Widget {
     void displayStock(int8_t displayIndex, StockDataModel &stock, uint32_t backgroundColor, uint32_t textColor);
 
     String m_apiKey = "c1tjb52ad3ia4h4uee9g";
-    unsigned long m_stockDelay = 60000;  //default to 60s between updates
+    unsigned long m_stockDelay = 300000;  //default to 5m between updates
     unsigned long m_stockDelayPrev = 0;
 
     StockDataModel m_stocks[MAX_STOCKS];
--- a/Info-Orbs/include/widgets/stockWidget.h
+++ b/Info-Orbs/include/widgets/stockWidget.h
@@ -22,7 +22,7 @@ class StockWidget : public Widget {
     void getStockData(StockDataModel &stock);
     void displayStock(int8_t displayIndex, StockDataModel &stock, uint32_t backgroundColor, uint32_t textColor);
 
-    String m_apiKey = "c1tjb52ad3ia4h4uee9g";
+    String m_apiKey = STOCK_TICKER_API_KEY;
     unsigned long m_stockDelay = 300000;  //default to 5m between updates
     unsigned long m_stockDelayPrev = 0;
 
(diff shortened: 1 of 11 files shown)


ericthelin:
This has a minor glitch that with the buttons being fully responsive with data at boot up the oddity of the pull down on the buttons gives a phantom "press" on boot of each button. This causes blinking to the weather app and back. We should make a fix to improve button handling to avoid that phantom click in a future PR.

Info-Orbs/include/widgets/stockWidget.h
```diff
@@ -1,38 +1,32 @@
...
-    void displayStock(int displayIndex, uint32_t backgroundColor, uint32_t textColor);
+   private:
+    void getStockData(StockDataModel &stock);
+    void displayStock(int8_t displayIndex, StockDataModel &stock, uint32_t backgroundColor, uint32_t textColor);
 
-    unsigned long m_stockDelay;
-    unsigned long m_stockDelayPrev;
-    int currentStockIndex;
-    std::vector<String> stockTickers;
+    String m_apiKey = "c1tjb52ad3ia4h4uee9g";
```
Jkitche, line 25: We should be careful about putting API keys in git - I've done this before and have royally horked the service it was for (spoiler: it was discord. The server owners were very unhappy with me)

ericthelin: Fair concern. I am just moving it here from being hardcoded into the url. This is Brett's call as he intends on paying for the API access and when I suggested that the weather api key should be a user configured value he wanted it hardcoded. So I split the difference here by making it easy to change but still hardcoded.

ericthelin: But I did go ahead and move it into the config.h.template like the weather api key. That way it is at least not in the code directly. Maybe we should talk Brett into distributing it outside of git.

tecnovel (approved):
looks good, have not fully studied the stockDataModel files but caching works good

tecnovel:
what are the refresh intervals right now here?

ericthelin:
Five minutes as requested by Brett in trello

tecnovel:
alright, i just noticed, when i fast switch through all the widgets, i get stuck in the weather widget for 1-2 seconds, probably because it needs to pull the api. but this does not happen in the stock api, why?

ericthelin:
It does happen with the existing dev branch. You are likely testing switching more than before. It has been getting in my way for a while. Take note of the blue led as it now works as a busy indicator.

ericthelin:
Now with catching working better I think we need to change the weather update interval.

ericthelin:
This is updated to the latest `dev` branch. As far as I know it is ready to merge.

ericthelin:
> alright, i just noticed, when i fast switch through all the widgets, i get stuck in the weather widget for 1-2 seconds, probably because it needs to pull the api. but this does not happen in the stock api, why?

I just looked at the weather widget and found the issue. I was resetting the update clock on widget switch. I fixed that and the delay you were seeing is fixed

tecnovel (approved):
now we are getting there! i like it!
can you make a short explanation what shouldInit() is doing? just for protocoll, after that i will merge it! so we have the new important functions somewhere documented. or are there other new important stuff you may want to add

tecnovel (approved):
love it!

PR #24, opened 2024-05-31
fix clock flicker for am/pm

ericthelin:
The clock does a full redraw each time so wiping the display is not needed.

### DIFF
--- a/Info-Orbs/lib/button/Button.cpp
+++ b/Info-Orbs/lib/button/Button.cpp
@@ -3,10 +3,17 @@
 
 
 Button::Button(uint8_t pin)
-    : m_pin(pin), m_delay(100), m_state(HIGH), m_ignoreUntil(0),
+    : m_pin(pin), m_delay(100), m_ignoreUntil(0),
       m_hasChanged(false) {}
 
-void Button::begin() { pinMode(m_pin, BUTTON_MODE); }
+void Button::begin() {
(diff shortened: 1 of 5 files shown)


Jkitche (approved):

tecnovel (changes requested):
IMG_1456

If you take out the display.fillScreen(TFT_BLACK); the screen has to be cleared somewhere else. (On the picture you see the first and last letter of my wifi name)

tecnovel (commented):
looks good(at least with refreshing the screens). but i found something weird, i just tested with these settings:
#define TIME_ZONE_OFFSET 2
#define FORMAT_24_HOUR false

An its 2:49(in the night) in my country but it shows 2:49 pm. did someone else a 24h test?

ericthelin:
yeah, the am/pm is broken at this point. in theory fixed in https://github.com/brettdottech/info-orbs/pull/25

ericthelin:
> looks good(at least with refreshing the screens). but i found something weird, i just tested with these settings: #define TIME_ZONE_OFFSET 2 #define FORMAT_24_HOUR false
> 
> An its 2:49(in the night) in my country but it shows 2:49 pm. did someone else a 24h test?

Jae's fixes are merged now. So it should be fixed now.

tecnovel (approved):
looks good

PR #25, opened 2024-06-01
Added setting for AM PM. Fixed 12 hour not being respected

Jkitche:
## Changes
This PR fixes the offset issue with the time provider always returning 12 when the time is past 12 pm. Instead, this switches to uses the libs `isPM` function instead.

Additionally, it adds a specific setting for showing the AM/PM indicator.

## Testing
Enable and disable the setting for the indicator to test the visibility
For very thorough testing, ensure the time is valid in both AM and PM scenarios. I will provide screenshots for both.

## Misc
image

### DIFF
--- a/Info-Orbs/src/widgets/clockWidget.cpp
+++ b/Info-Orbs/src/widgets/clockWidget.cpp
@@ -17,7 +17,6 @@ void ClockWidget::setup() {
 }
 
 void ClockWidget::draw(bool force) {
-    GlobalTime* time = GlobalTime::getInstance();
     bool showFirstDigit = FORMAT_24_HOUR || (!FORMAT_24_HOUR && m_hourSingle >= 10);
     if ((showFirstDigit && m_lastDisplay1Didget != m_display1Didget) || force) {
         displayDidget(0, m_display1Didget, 7, 5, FOREGROUND_COLOR);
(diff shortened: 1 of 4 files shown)


Info-Orbs/src/widgets/clockWidget.cpp
```diff
@@ -17,7 +17,8 @@ void ClockWidget::setup() {
 }
 
 void ClockWidget::draw(bool force) {
-    bool showFirstDigit = FORMAT_24_HOUR || (!FORMAT_24_HOUR && m_hourSingle >= 10);
+    GlobalTime* time = GlobalTime::getInstance();
+    bool showFirstDigit = FORMAT_24_HOUR || (!FORMAT_24_HOUR && time->isPM());
```
ericthelin, line 21: this is wrong. I think the first version is correct

Info-Orbs/src/widgets/clockWidget.cpp
```diff
@@ -49,11 +50,12 @@ void ClockWidget::draw(bool force) {
 }
 
 void ClockWidget::displayAmPm(uint32_t color) {
+    GlobalTime* time = GlobalTime::getInstance();
     m_manager.selectScreen(2);
     TFT_eSPI& display = m_manager.getDisplay();
     display.setTextSize(4);
     display.setTextColor(color, TFT_BLACK, true);
-    const String& am_pm = m_hourSingle >= 12 ? "AM" : "PM";
+    const String& am_pm = time->isPM() ? "PM" : "AM";
```
ericthelin, line 58: why `const String &` instead of just `String`

Jkitche: It was just to match the types of `display.drawString`

Info-Orbs/src/widgets/clockWidget.cpp
Jkitche: Yup, you're right. Got my changes mixed up.

ericthelin (approved):

ericthelin:
Tested with all cases that seem relevant (24h, 12h pm, 12h am, 12h not shown)

PR #26, opened 2024-06-01
show shadow for single digit hours and clear for non-shadow

ericthelin:
This fixes a bug where the transition from a double digit hour to a single digit hour did not clear the first digit. and restores the shadow on the first digit.

This also adds a call to ScreenManager to clear a single display.

### DIFF
--- a/Info-Orbs/src/widgets/clockWidget.cpp
+++ b/Info-Orbs/src/widgets/clockWidget.cpp
@@ -75,7 +75,7 @@ void ClockWidget::update(bool force) {
         if (m_hourSingle < 10) {
             m_display1Didget = " ";
         } else {
-            m_display1Didget = "1";
+            m_display1Didget = int(m_hourSingle/10);
         }
         m_display2Didget = m_hourSingle % 10;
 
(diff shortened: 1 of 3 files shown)


Info-Orbs/src/widgets/clockWidget.cpp
```diff
@@ -65,14 +67,17 @@ void ClockWidget::update(bool force) {
...
-        String currentHourPadded = String(m_hourSingle).length() == 1 ? " " + String(m_hourSingle) : String(m_hourSingle);
-
-        m_display1Didget = currentHourPadded.substring(0, 1);
-        m_display2Didget = currentHourPadded.substring(1, 2);
+        if (m_hourSingle < 10) {
+            m_display1Didget = " ";
+        } else {
+            m_display1Didget = "1";
+        }
+        m_display2Didget = m_hourSingle % 10;
```
tecnovel, line 80: this will not work for 24 hours clock. First digit also can be a 2 ;)

tecnovel (changes requested):

Info-Orbs/src/widgets/clockWidget.cpp
ericthelin: good catch. Fixed and tested.

tecnovel (approved):
thanks ;)

PR #27, opened 2024-06-03
timezone API

tecnovel:
**please consider the changes in config.h.template!**

Trello Card: https://trello.com/c/xjwMScgJ/79-get-timezone-offset-on-boot-from-api

## Changes
the globalTime class get's now the timezone offset via a seperate api so it is not depending on other widgets adn you can configure a seperate location for weather and time if you want to

### DIFF
--- a/Info-Orbs/lib/globalTime/globalTime.cpp
+++ b/Info-Orbs/lib/globalTime/globalTime.cpp
@@ -9,7 +9,6 @@ GlobalTime::GlobalTime() {
     m_timeClient = new NTPClient(m_udp);
     m_timeClient->begin();
     m_timeClient->setPoolServerName(NTP_SERVER);
-    //m_timeClient->setTimeOffset(3600 * (TIME_ZONE_OFFSET));
 }
 
 GlobalTime::~GlobalTime() {
@@ -25,8 +24,7 @@ GlobalTime *GlobalTime::getInstance() {
 
 void GlobalTime::updateTime() {
     if (millis() - m_updateTimer > m_oneSecond) {
-        if (m_timeZoneOffset == 1000) {
-            //m_timeClient->setTimeOffset(3600 * m_timeZoneOffset);
+        if (m_timeZoneOffset == -1) {
             getTimeZoneOffsetFromAPI();
         }
         m_timeClient->update();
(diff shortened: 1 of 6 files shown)


Info-Orbs/lib/globalTime/globalTime.cpp
```diff
@@ -93,12 +94,29 @@ String GlobalTime::getWeekday() {
...
+    if (httpCode > 0) {
+        JsonDocument doc;
+        DeserializationError error = deserializeJson(doc, http.getString());
+        if (!error) {
+            m_timeZoneOffset = doc["gmtOffset"].as<int>();
+            Serial.print("Timezone Offset from API: ");
+            Serial.println(m_timeZoneOffset);
+            m_timeClient->setTimeOffset(m_timeZoneOffset);
+        } else {
+            Serial.println("Failed to get timezone offset from API");
```
ericthelin, line 117: the error should mention deserialization or something to be distinct from the http error

Info-Orbs/lib/globalTime/globalTime.cpp
```diff
@@ -9,7 +9,7 @@ GlobalTime::GlobalTime() {
     m_timeClient = new NTPClient(m_udp);
     m_timeClient->begin();
     m_timeClient->setPoolServerName(NTP_SERVER);
-    m_timeClient->setTimeOffset(3600 * (TIME_ZONE_OFFSET));
+    //m_timeClient->setTimeOffset(3600 * (TIME_ZONE_OFFSET));
```
ericthelin, line 12: I don't think we need to keep the commented version around so just remove it

Info-Orbs/lib/globalTime/globalTime.cpp
```diff
@@ -25,8 +25,9 @@ GlobalTime *GlobalTime::getInstance() {
 
 void GlobalTime::updateTime() {
     if (millis() - m_updateTimer > m_oneSecond) {
-        if (m_timeZoneOffset != 0) {
-            m_timeClient->setTimeOffset(3600 * m_timeZoneOffset);
+        if (m_timeZoneOffset == 1000) {
```
ericthelin, line 28: maybe a default value of "-1" or a constant to indicate unset?

tecnovel: ahh, i thought -1 does not work because timezone can also be negative, but the response is already multiplied by 3600 so it should work

Info-Orbs/lib/globalTime/globalTime.cpp
tecnovel: yeah sure, cp mistake

PR #28, opened 2024-06-06
Weather refactor

ericthelin:
This improves the code in lots of small ways. But it also brings in high and low temperature displays on both the today orb and the forecast orb.

### DIFF
--- a/Info-Orbs/include/widgets/stockDataModel.h
+++ b/Info-Orbs/include/widgets/stockDataModel.h
@@ -1,0 +1,51 @@
+#ifndef STOCK_DATA_MODEL_H
+#define STOCK_DATA_MODEL_H
+
+#include <Arduino.h>
+
+#include <iomanip>
+#include <sstream>
+
+class StockDataModel {
+   public:
+    StockDataModel();
+    StockDataModel &setSymbol(String symbol);
+    String getSymbol();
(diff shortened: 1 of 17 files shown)


ericthelin:
There will be one more refactor at some point to introduce a data model class. But that can wait and be a separate PR.

tecnovel (changes requested):
Do I need to change something in the config.h? no right? because i have some weir behavior here.
Mittel (IMG_1492)
Mittel (IMG_1493)
Mittel (IMG_1494)

- The Stock Widget showed only one stock, and now after a while waiting the others where loaded too. 
- in the weather widget i have an error printed out: CertificatePoliciesJ... (should be the date there)
- On the big temperature orb, can you add a space between the number and the ° ?

tecnovel:
I haven't looked into the code and did not debug anything. I don't have time at the moment, sorry, but i will later this weekend. but maybe you can solve it already until then...

ericthelin:
for the certificate issue that is well outside of these changes. still an issue but unrelated.
for the stocks there isn't any retry logic on failures it just skips the update for that stock. Which is unfortunate when it is the initial load.
I will add a little more space before the degree symbol. Finding the right distance is a little tricky now because it is autocalculated.

tecnovel (approved):
okey, let's do the other stuff in another PR

PR #31, opened 2024-06-10
small fixes and cleanup

tecnovel:
## Changes
just a clean up and small changes:

- if you switch between the widgets, the weather widget now waits 1sec before pulling the data from api so you can swicth without waiting for the API


tecnovel:
should we add this too to stock widget? i saw the timing is now handled completely different...
just a proposal to discuss

ericthelin (approved):

ericthelin:
yes, this is a good concept and should be added to stocks. The approach is much more similar than it seems

tecnovel:
i have it now implemented also in the stock widget.
so it waits now 1s before refreshing the data when you change the widget

tecnovel:
@ericthelin  i've merged the updates from dev into this one and now i've got sometimes a really weird behavior.
when i download the code and want to switch a widget, it shows no data, the it loads it and cache it, then everything is okey. but it looks like it cached it not while init. and then i wanted to debug with serial monitor, and it shows me weird charactes sometimes (most of the time it works, but sometimes it prints ÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿ.

do you also have this? and any idea why? before i merged the dev into this branch everything was okey, and on the dev branch i don't see this behavior

ericthelin:
The printing strange characters is normal from my experience. I blame timing of the serial monitor connecting. My theory is that the output is all bit shifted into that from the monitor getting connected in a strange part of the bit stream. No evidence to support that theory but that is the best I have. And I know that it ALWAYS works after resetting the device when that happens. So I haven't worried about it.

PR #32, opened 2024-06-10
proposal for errorhandling

tecnovel:
Trello Card: https://trello.com/c/MtMIv0r7/83-stock-bug

## Changes
try to figure out a proper error handling for widgets


brammie15:
Hmmm, it's a good idea. 


Hear me out, we have a singleton (i know i know bad bad >:( 
But this singleton can "override" the drawing commands as such and basically draw over the screen. This would have to be a "flag" of some sort in the main loop to stop whatever the current widget is doing and possibly reset the currentWidget aswell

learn-more:
This should probably not be a PR but a comment in the discord or a discussion item?
@brett-dot-tech would it make sense to close this?

PR #34, opened 2024-06-15
refactor weather widget to use a data model

ericthelin:


brammie15:
Code looking good!

Gonna wait for some1 else to approve but looks good (Will give it a quick test :p)

tecnovel (changes requested):
add MAX_RETRIES define to the config

when you push the button, it should change at the 3days forecast high and low temp, but it does not. it looks right in the weatherwidget.cpp so the bug has to be when loading the data

tecnovel (approved):

tecnovel:
@ericthelin do you accept my bugfix? (i can not request a review from you here)

ericthelin:
> @ericthelin do you accept my bugfix? (i can not request a review from you here)

Good catch

PR #38, opened 2024-07-05
selectable 24 hour clock format

JulianSelf:
added property to globalTime to reflect the 24 hour format, initialised from FORMAT_24_HOUR
added new define SHOW_SECOND_TICKS to drive showing second tick marks going around the pulsing dots in the middle display
updated clockWidget change method to toggle 24 hour format


Info-Orbs/.vscode/settings.json
tecnovel, line 1: pleaase delete this change

tecnovel (commented):
why do you want to change the 24h mode on runtime? just a general question...

Info-Orbs/.vscode/settings.json
JulianSelf: Is that a quick way to do that?
(Or happy for you to do it if that is quicker?)

tecnovel (approved):

PR #44, opened 2024-09-06
INVERTED_ORBS config option

racingsquirrel:
Add config option to invert the Orbs. Inverts the screens and re-orders screens and buttons.


ericthelin (approved):

brammie15:
Dont like this. Would rather see a setting on the display Manager that just inverts the display it writes to
```cpp
If(m_inverted) {
   CurrentIndex = 5 - i;
} Else {
   CurrentIndex = i;
}
```

Since i am not a fan of changing the config

brett-dot-tech:
Making a note here that we will work to change this to implement in a cleaner way however merging so the functionality is there for those who need it (:

PR #46, opened 2024-09-09
Added Gerber FIles

LeDomme:
Added Gerber files for easier order, especially for people without KiCad knowledge


LeDomme:
Missed stuff, PR came to quick, will do another

PR #49, opened 2024-09-12
Widget Selector

LeDomme:
added a simple method for the user to select which widget to activate.
also reorganized the config menu so settings with the same topic stay together.

and yes you can deactivate everything and enter a bootloop, nothing that can't be fixed by correcting the config and reflashing.

### DIFF
--- a/Info-Orbs/include/widgets/stockWidget.h
+++ b/Info-Orbs/include/widgets/stockWidget.h
@@ -1,3 +1,4 @@
+
 #ifndef STOCK_WIDGET_H
 #define STOCK_WIDGET_H
 
--- a/Info-Orbs/include/widgets/weatherWidget.h
+++ b/Info-Orbs/include/widgets/weatherWidget.h
@@ -1,3 +1,4 @@
+#ifdef WEATHER_LOCAION
 #ifndef WEAHTERWIDGET_H
 #define WEAHTERWIDGET_H
 
@@ -61,3 +62,4 @@ class WeatherWidget : public Widget {
     const int MODE_LOWS = 1;
 };
 #endif  // WEAHTERWIDGET_H
+#endif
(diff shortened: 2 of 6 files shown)


ericthelin (commented):
I like the concept. But rather than introducing three more configs that people need to copy to their config.h what if we used WEATHER_LOCATION and STOCK_TICKER_LIST instead? And maybe even reverse the clock value to IGNORE_CLOCK (or something like that) so that if people don't update their config at all they don't have any change in functionality? This is all similar to the WEB_DATA_WIDGET_URL being the trigger to determine if that module is intialized. We can then also put #ifdef around the entire modules and if you turn them off they don't even get compiled onto the device. That way if we run out of memory which seems to be imminent we can simply choose one or more widget to exclude.

LeDomme:
Changed it to use the WEATHER_LOCATION and STOCK_TICKER_LIST.

Info-Orbs/include/widgets/stockWidget.h
```diff
@@ -29,4 +30,4 @@ class StockWidget : public Widget {
     StockDataModel m_stocks[MAX_STOCKS];
     int8_t m_stockCount;
 };
-#endif  // STOCK_WIDGET_H
+#endif  // STOCK_WIDGET_H
```
learn-more, line 33: Removing final newlines from a file is no improvement.

brett-dot-tech:
@LeDomme I am SO sorry about the delay here dude. Just getting my feet under me with PRs here. I really like this and its actually something on the top of my list to roll out personally, so you killed this. 

when you have a sec, could you instead make this PR in the dev branch? (there may be some small tweaks required as well due to some small chnages in the config file). No rush at all, I appericate you on this one.

brett-dot-tech:
Closing this as the codebase has strayed far from this being mergable, feel free to open again though (:

PR #50, opened 2024-09-15
Upgraded web data widget

ericthelin:
Add more options to generate a complex image from a JSON webserver.

PXL_20240918_015638548

Here is my test php script that shows a demo of the new options:

```<?php
$sensorsStr = shell_exec('/usr/bin/sensors -j');
$sensors = json_decode($sensorsStr, true);

$uptimeOutput = shell_exec('uptime');
if (preg_match('/^\s*(..:..:..) up (.*),  ([0-9]+ users),  load average: ([0-9.]+), ([0-9.]+), ([0-9.]+)$/', $uptimeOutput, $output)) {
	$time = $output[1];
	$uptime = preg_replace('/, /', " ", $output[2]);
	$load1min = $output[4];
	$load5min = $output[5];
	$load15min = $output[6];
}

$uptime = "line 1\nline 2 longer line\nline 3\n";
$uptime = [
	[
		"type" => "circle",
		"x" => 50,
		"y" => 50,
		"radius" => 20,
		"color" => "red",
		"filled" => true,
	],
	[
		"type" => "text",
		"x" => 120,
		"y" => 170,
		"font" => 1,
		"text" => "TestMe ",
		"color" => "white",
		"background" => "darkgrey",
	],
	[
		"type" => "rectangle",
		"x" => 70,
		"y" => 180,
		"height" => 50,
		"width" => 100,
		"filled" => true,
		"color" => "blue"
	],
	[
		"type" => "triangle",
		"x" => 120,
		"y" => 120,
		"x2" => 180,
		"y2" => 60,
		"x3" => 60,
		"y3" => 60,
		"filled" => false,
		"color" => "red"
	],
	[
		"type" => "line",
		"x" => 60,
		"y" => 120,
		"x2" => 180,
		"y2" => 120,
		"color" => "blue"
	],
	[
		"type" => "arc",
		"x" => 120,
		"y" => 120,
		"radius" => 30,
		"innerRadius" => 20,
		"angleStart" => 270,
		"angleEnd" => 90,
		"color" => "blue"
	],
	[
		"type" => "character",
		"x" => 120,
		"y" => 190,
		"character" => "E",
		"color" => "green",
		"font" => 2,
	],
];


$output = [
	[
		'label' => 'CPU',
		'data' => $sensors['coretemp-isa-0000']['Package id 0']['temp1_input'],
		'labelColor' => 'blue',
		'color' => 'red',
		'background' => 'silver',
	],
	[
		'label' => 'PNY NVME',
		'data' => $sensors['nvme-pci-0200']['Composite']['temp1_input'],
	],
	[
		'label' => 'WD Blue NVME',
		'data' => $sensors['nvme-pci-0300']['Composite']['temp1_input'],
	],
	[
		'label' => 'Uptime',
		'data' => $uptime,
		'labelX' => 100,
		'labelY' => 100,
	],
	[
		'label' => 'Load 1 min',
		'data' => $load1min,
		'color' => 'green',
		'labelColor' => 'dArK Green',
	],
];

header('Content-Type: application/json; charset=utf-8');

// print '[{"label":"CPU","data":36},{"label":"PNY NVME","data":45.85},{"label":"WD Blue NVME","data":37.85},{"label":"Uptime","data":"19 days 21:59","labelColor":"blue","color":"red","background":"silver"},{"label":"Load 1 min","data":"0.79","color":"green","labelColor":"dArK Green"}]';
// exit;
print json_encode($output) . "\n";
```

And to match the existing stock display use
```<?php
$stocks = explode(",", $_GET['stocks']);
$displays = ['interval' => '20000', 'displays' => []];
foreach ($stocks as $stock) {
	$stockData = outputStockDisplay($stock);
	$displays['displays'][] = $stockData;
}

header('Content-Type: application/json; charset=utf-8');
echo json_encode($displays);
exit;

function get_stock_data($symbol)
{
	$apiKey = 'aVhwT1NWWkhIZVBRZlIwOUlHb01keWFrMEI5Ql9QM1ZIZndtay1ub0V3OD0';
	$url = sprintf('https://api.marketdata.app/v1/stocks/quotes/%s/?token=%s', $symbol, $apiKey);
	$json = file_get_contents($url);
	return json_decode($json, true);
}

function outputStockDisplay($symbol)
{
	$data = get_stock_data($symbol);
	$price = $data['last'][0];
	$change = $data['change'][0];
	$changePercent = $data['changepct'][0];
	$changePercent = round($changePercent * 100, 2);

	$headerColor = 'blue';

	$output = [
		'fullDraw' => false,
		'data' => [
			[
				'type' => 'rectangle',
				'x' => 0,
				'y' => 0,
				'height' => 50,
				'width' => 240,
				'color' => $headerColor,
				'filled' => true,
			],
			[
				'type' => 'text',
				'x' => 120,
				'y' => 27,
				'font' => 1,
				'size' => 4,
				'text' => $symbol,
				'align' => 'center',
				'color' => 'white',
				'background' => $headerColor,
			],
			[
				'type' => 'text',
				'x' => 120,
				'y' => 51 + 32,
				'font' => 1,
				'size' => 4,
				'align' => 'center',
				'text' => sprintf('$%.2f', $price),
				'color' => 'white',
				'background' => 'black',
			],
			[
				'type' => 'text',
				'x' => 120,
				'y' => 147,
				'font' => 1,
				'size' => 4,
				'align' => 'center',
				'text' => sprintf('%.2f%%', $changePercent),
				'color' => 'green',
				'background' => 'black',
			],
		]
	];

	if ($change > 0) {
		$output['data'][] = [
			'type' => 'triangle',
			'x' => 120,
			'y' => 185,
			'x2' => 140,
			'y2' => 220,
			'x3' => 100,
			'y3' => 220,
			'filled' => true,
			'color' => 'green',
		];
	} else {
		$output['data'][] = [
			'type' => 'triangle',
			'x' => 120,
			'y' => 220,
			'x2' => 140,
			'y2' => 185,
			'x3' => 100,
			'y3' => 185,
			'filled' => true,
			'color' => 'red',
		];
	}
	return $output;
}
```


Info-Orbs/src/widgets/webDataElementModel.cpp
```diff
@@ -0,0 +1,316 @@
...
+String WebDataElementModel::getType() {
+    return m_type;
+}
+
+void WebDataElementModel::setType(String type) {
+    if (m_type != type) {
+        m_type = type;
+
+        if (m_type == "text") {
+            m_typeEnum = TEXT;
```
learn-more, line 12: It might make sense to make dedicated `TextWebDataElement` items, (and all other types) so they can read their own properties, and only contain the relevant fields?

Info-Orbs/src/widgets/webDataModel.cpp
```diff
@@ -18,10 +18,68 @@ String WebDataModel::getData() {
...
+    return m_elements[index];
+}
+
+int32_t WebDataModel::getElementsCount() {
+    return m_elementsCount;
+}
+
+void WebDataModel::setElementsCount(int32_t count) {
+    if (m_elementsCount != count) {
+        m_elements = (WebDataElementModel *)malloc(sizeof(WebDataElementModel) * count);
```
learn-more, line 70: This does not run the constructor, so it might be safer to do `m_elements = new WebDataElementModel[count];`

learn-more: Additionally, if `m_elements` is already set here, you leak that so you need to `free(m_elements)` right now,
but if you go for the `new[]` then you need to `delete[] m_elements;`

Info-Orbs/src/widgets/webDataWidget.cpp
```diff
@@ -34,12 +34,56 @@ void WebDataWidget::draw(bool force) {
...
-            String wrappedLines[MAX_WRAPPED_LINES];
-            String dataValues = data->getData();
-            int lineCount = Utils::getWrappedLines(wrappedLines, dataValues, 10);
-            int height = display.fontHeight() + 10;
-            for (int i = 0; i < lineCount; i++) {
-                display.drawString(wrappedLines[i], 120, yOffset + (height*i), 2);
+            if (data->getElementsCount() > 0) {
+                for (int i = 0; i < data->getElementsCount(); i++) {
+                    WebDataElementModel element = data->getElement(i);
+                    switch (element.getTypeEnum()) {
```
learn-more, line 40: If you create a type per item, then this can just be `element->draw()` or something like that,
and the item itself can handle the drawing.

Info-Orbs/src/widgets/webDataWidget.cpp
```diff
@@ -34,12 +34,56 @@ void WebDataWidget::draw(bool force) {
...
 
-            String wrappedLines[MAX_WRAPPED_LINES];
-            String dataValues = data->getData();
-            int lineCount = Utils::getWrappedLines(wrappedLines, dataValues, 10);
-            int height = display.fontHeight() + 10;
-            for (int i = 0; i < lineCount; i++) {
-                display.drawString(wrappedLines[i], 120, yOffset + (height*i), 2);
+            if (data->getElementsCount() > 0) {
+                for (int i = 0; i < data->getElementsCount(); i++) {
+                    WebDataElementModel element = data->getElement(i);
```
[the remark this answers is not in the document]
ericthelin: fixed.

learn-more (commented):
I left a bunch of notes / improvements,
most 'new concepts' that I introduced are only commented once, not every instance where they could be applied.
Missing initializers and stuff like that are commented everywhere I noticed them.

Info-Orbs/src/widgets/webDataModel.cpp
```diff
@@ -18,6 +18,39 @@ String WebDataModel::getData() {
...
         m_data = data;
+        if (data[0] == '[') {
+            Serial.println("JSON");
+            JsonDocument doc;
+            DeserializationError error = deserializeJson(doc, data);
+            setElementsCount(doc.size());
+            for (int i = 0; i < doc.size(); i++) {
+                WebDataElementModel element = WebDataElementModel();
+                element.parseData(doc[i].as<JsonObject>());
+                m_elements[i] = element;
```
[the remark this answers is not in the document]
ericthelin: I must be missing something as this doesn't compile for me.

Info-Orbs/include/widgets/webDataElementCharacterModel.h
```diff
@@ -27,15 +27,15 @@ class WebDataElementCharacterModel: public WebDataElement {
...
-    void parseData(JsonObject doc);
-    void draw(TFT_eSPI& display, int32_t defaultColor, int32_t defaultBackground);
+    void parseData(const JsonObject& doc, int32_t defaultColor, int32_t defaultBackground) override;
+    void draw(TFT_eSPI& display) override;
 
    private:
     int32_t m_x = 0;
     int32_t m_y = 0;
-    String m_character;
+    String m_character = "";
```
learn-more, line 36: A `String` is an object that should self-initialize, so this is the exception that does not need to be initialized here.

Info-Orbs/include/widgets/webDataElement.h
```diff
@@ -0,0 +1,21 @@
+#ifndef WEB_DATA_ELEMENT_H
+#define WEB_DATA_ELEMENT_H
+
+#include <ArduinoJson.h>
+#include <TFT_eSPI.h>
+
+class WebDataElement {
```
learn-more, line 7: It would probably make sense to move this class (and all things derived from it) into a folder `WebData` or something,
because this is an implementation detail of the `WebDataModel` and not a widget.

Info-Orbs/src/widgets/webDataModel.cpp
```diff
@@ -70,6 +103,59 @@ void WebDataModel::setBackgroundColor(String background) {
...
+
+void WebDataModel::parseData(const JsonObject& doc, int32_t defaultColor, int32_t defaultBackground) {
+    if (doc.containsKey("label")) {
+        setLabel(doc["label"].as<String>());
+    }
+    if (doc.containsKey("data")) {
+        setData(doc["data"].as<String>(), defaultColor, defaultBackground);
+    }
+    if (doc.containsKey("color")) {
+        setDataColor(doc["color"].as<String>());
```
learn-more, line 119: I don't think you need the `.as<String>()` for this library.

ericthelin: yes, we do `doc["color"]` is a JsonObject so we need `.as<String>()` to convert it to a string. I suppose we could pass the JsonObject and cast it in the function but I don't see any benefits to that.

Info-Orbs/include/widgets/webDataElement.h
ericthelin: I moved all data models to a `model` directory

Info-Orbs/src/widgets/webDataModel.cpp
learn-more: The documentation on https://arduinojson.org/ claims:
> Implicit or Explicit Casts
> ArduinoJson supports two coding styles, with implicit or explicit casts
> 
> Implicit cast:
>
> ```
> const char* city = doc["city"];
> float temp = doc["temperature"];
> ```
> Explicit cast:
>
> ```
> auto city = doc["city"].as<const char*>();
> auto temp = doc["temperature"].as<float>();
> ```

ericthelin: I think that would work if we copied it to a variable but using it as an argument for a function call doesn't seem to be enough context for the implicit casts.

Info-Orbs/include/widgets/webDataWidget.h
```diff
@@ -1,28 +1,28 @@
...
+   public:
     WebDataWidget(ScreenManager &manager, String url);
     ~WebDataWidget() override;
     void setup() override;
     void update(bool force = false) override;
     void draw(bool force = false) override;
     void changeMode() override;
 
-private:
+   private:
```
learn-more, line 20: Is this indenting done by a `.clang-format` file or something like it?
I am not seeing one in the project, so it might be an idea to add one in the foreseeable future.
This helps contributors maintain the same style.

Info-Orbs/src/core/utils.cpp
```diff
@@ -1,6 +1,5 @@
 #include "utils.h"
 
-
 int Utils::getWrappedLines(String (&lines)[MAX_WRAPPED_LINES], String str, int limit) {
```
learn-more, line 3: It would be good to start thinking about unit-tests for code like this.

Info-Orbs/include/widgets/webDataWidget.h
learn-more: ps: I think this indenting style is fairly weird, with the access specifier indented one space less than the rest.

Info-Orbs/src/model/webDataModel.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+String WebDataModel::getData() {
+    return m_data;
+}
+
+void WebDataModel::setData(String data, int32_t defaultColor, int32_t defaultBackground) {
+    if (m_data != data) {
+        m_data = data;
+        if (data[0] == '[') {
+            // Serial.println("JSON");
```
learn-more, line 22: Debugging leftover?
(there are a lot more, but I won't point them all out individually)

Info-Orbs/src/model/webDataModel.cpp
```diff
@@ -0,0 +1,188 @@
...
+            JsonDocument doc;
+            DeserializationError error = deserializeJson(doc, data);
+            setElementsCount(doc.size());
+            for (int i = 0; i < doc.size(); i++) {
+                WebDataElementModel element = WebDataElementModel();
+                element.parseData(doc[i].as<JsonObject>(), defaultColor, defaultBackground);
+                m_elements[i] = element;
```
learn-more, line 29:
```suggestion
                m_elements[i].parseData(doc[i].as<JsonObject>(), defaultColor, defaultBackground);
```

Info-Orbs/src/model/webDataModel.cpp
```diff
@@ -0,0 +1,188 @@
...
+            }
+        } else {
+            setElementsCount(0);
+        }
+        m_changed = true;
```
learn-more, line 34: If we keep an `m_changed` at this level, the individual elements don't really need it,
and the getter / setters could be removed?

learn-more (approved):
There are some points that can be improved, but it might be a good idea to move forward with this as to not keep this PR open for months on end.

Info-Orbs/include/widgets/webDataWidget.h
ericthelin: I know it was in the project config at one point but I am not seeing it anymore. This is what I have in my VS Code settings.json
`"C_Cpp.clang_format_fallbackStyle": "{ BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 0}",`
I agree it seems weird. I am just going with it because it seems to be a standard.

Info-Orbs/src/core/utils.cpp
ericthelin: I haven't written unit tests for C++ code but I agree they would be useful.

Info-Orbs/src/model/webDataElementArcModel.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+int32_t WebDataElementArcModel::getColor() {
+    return m_color;
+}
+
+void WebDataElementArcModel::parseData(const JsonObject& doc, int32_t defaultColor, int32_t defaultBackground) {
+    // Serial.println("Parsing Data");
+
+    if (doc.containsKey("x")) {
```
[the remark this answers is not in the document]
ericthelin: I agreee. What do they recommend replacing it with?

Info-Orbs/src/model/webDataElementImageModel.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+    if (doc.containsKey("image")) {
+        setImage(doc["image"].as<String>());
+    }
+    // if (doc.containsKey("imageUrl")) {
+    //     setImageUrl(doc["imageUrl"].as<String>());
+    // }
+    // if (doc.containsKey("imageBytes")) {
+    //     setImageBytes(doc["imageBytes"].as<String>());
+    // }
```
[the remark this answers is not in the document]
ericthelin: more example of what I am expecting to have in the near future. As this entire class is currently a placeholder.

Info-Orbs/src/model/webDataModel.cpp
ericthelin: yes, that m_changes should be moved into the else above as it is intended for the old version where data is a single value.

Info-Orbs/src/model/webDataElementArcModel.cpp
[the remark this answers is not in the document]
learn-more: https://arduinojson.org/v7/api/jsondocument/containskey/
For int and the like, you can use `is<int>()`,
for strings you can just read it and it will be a `nullptr` if there is no value:
```cpp
const char* city = doc["city"];
if (city)
  Serial.println(city);
```

For example, reading the color previously:
```cpp
if (doc[i].containsKey("color")) {
    data->setDataColor(doc[i]["color"].as<String>());
```
Could be written as:
```cpp
if (const char* color = doc[i]["color"]) {
    data->setDataColor(color);
```

Info-Orbs/src/model/webDataModel.cpp
ericthelin: but on second thought maybe I need to put some more thought into changed state with regard to the individual element models. As I don't think it is being used right now but it should be. For now I am going to leave this in place and maybe it will stay but I will add checks on the draw method for each element too. So this will mark the wholeDataModel as in need of redrawing which means it will look at each individual element to determine if it needs to be redrawn.Or maybe we will stick to redrawing the entire screen if there any changes. Either way it should be resolved in a future PR.

Info-Orbs/src/model/webDataModel.cpp
ericthelin: done

Info-Orbs/src/model/webDataElementArcModel.cpp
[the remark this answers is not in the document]
ericthelin: Using is<int>() is fine. But creating a variable that is only used for a function call seems like a bad idea. But we are using arduinojson v7 already. So I suppose it needs to be done.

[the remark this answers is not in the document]
ericthelin: I have started making this change but strings will need more testing. So it will be a followup PR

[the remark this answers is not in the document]
learn-more: > Using is() is fine. But creating a variable that is only used for a function call seems like a bad idea. But we are using arduinojson v7 already. So I suppose it needs to be done.


> But creating a variable that is only used for a function call seems like a bad idea. 

It is not, in this case it only has to do the lookup once instead of twice so overall it will be _way_ cheaper.

brett-dot-tech (approved):

PR #52, opened 2024-09-19
Embed jpg files using the pio options instead of hand-coding them

learn-more:
The syntax is a bit 'interesting',
but it allows having the files as .jpg in the source tree.

Relevant part from icons.h (github hides it by default):
```cpp
// These symbols are generated from the files specified in platformio.ini under 'board_build.embed_files'
// See https://docs.platformio.org/en/latest/platforms/espressif32.html#embedding-binary-data for more info

extern const byte moonCloud_start[] asm("_binary_icons_moonCloud_jpg_start");
extern const byte moonCloud_end[] asm("_binary_icons_moonCloud_jpg_end");
extern const byte sunClouds_start[] asm("_binary_icons_sunClouds_jpg_start");
extern const byte sunClouds_end[] asm("_binary_icons_sunClouds_jpg_end");
extern const byte sun_start[] asm("_binary_icons_sun_jpg_start");
extern const byte sun_end[] asm("_binary_icons_sun_jpg_end");
extern const byte moon_start[] asm("_binary_icons_moon_jpg_start");
extern const byte moon_end[] asm("_binary_icons_moon_jpg_end");
extern const byte snow_start[] asm("_binary_icons_snow_jpg_start");
extern const byte snow_end[] asm("_binary_icons_snow_jpg_end");
extern const byte rain_start[] asm("_binary_icons_rain_jpg_start");
extern const byte rain_end[] asm("_binary_icons_rain_jpg_end");
extern const byte clouds_start[] asm("_binary_icons_clouds_jpg_start");
extern const byte clouds_end[] asm("_binary_icons_clouds_jpg_end");
```

-- This is a draft because I have not been able to test this on a device yet --


ericthelin:
It does build. But it seems to still use regular storage.
```Flash: [==========]  97.5% (used 1277637 bytes from 1310720 bytes)```

learn-more:
> It does build. But it seems to still use regular storage. `Flash: [==========] 97.5% (used 1277637 bytes from 1310720 bytes)`

Indeed, that was not the goal of this PR.

The main 'win' from this PR is that it allows users to edit the jpg files without having to re-create the in-source hexdump.

ericthelin (approved):

PR #53, opened 2024-09-19
Remove usage of stringstreams to reduce the binary size

learn-more:
Before:
```
RAM:   [==        ]  17.1% (used 56012 bytes from 327680 bytes)
Flash: [==========]  95.2% (used 1248121 bytes from 1310720 bytes)
```

After:
```
RAM:   [==        ]  15.6% (used 51140 bytes from 327680 bytes)
Flash: [========  ]  78.6% (used 1030505 bytes from 1310720 bytes)
```

-- This is a draft because I have not been able to test this on a device yet --


ericthelin:
This does seem to work! But strangely I am not getting quite the same storage used. But absolutely worth using.
```Flash: [========  ]  80.9% (used 1059745 bytes from 1310720 bytes)```

ericthelin (approved):

PR #54, opened 2024-09-19
Create workflow that builds the PlatformIO project

learn-more:
Every PR and every push to main will be built.

### DIFF
--- a/.github/workflows/platformio.yml
+++ b/.github/workflows/platformio.yml
@@ -1,0 +1,36 @@
+name: PlatformIO CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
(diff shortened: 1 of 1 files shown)


ericthelin (approved):
I like the idea. I think the only way to test it is to commit it and make a new change if it doesn't work, correct?

learn-more:
> I like the idea. I think the only way to test it is to commit it and make a new change if it doesn't work, correct?

In general yes.

Usually the workflow is:
- User creates some code
- User pushes this to his own branch (he could enable a CI there, but by default it's only enabled on main to prevent lots of builder time waste).
- User creates a PR (now it will be built automatically)
  - Build fails, user can fix it and push to his local branch again
  - This triggers a re-build in the PR
- Changes merged to main, this triggers a build as well.

PR #64, opened 2024-10-04
Spelling (didget -> digit), also Widgets now have a getName() method

RobThree:
<sup>[3rd](https://github.com/brettdottech/info-orbs/pull/62) [time](https://github.com/brettdottech/info-orbs/pull/63)'s the charm 😅 Sorry, my bad!</sup>

Unless this was intentional(?), it's spelled digit 😉

Also Widgets now have a `getName()` method so the name of the currently loading widget can be displayed during startup (both on serial and the displays).

I do realize this should've been 2 separate PR's but I'm hoping this is fine.

### DIFF
--- a/Info-Orbs/include/core/wifiWidget.h
+++ b/Info-Orbs/include/core/wifiWidget.h
@@ -1,7 +1,8 @@
 #pragma once
 #ifndef WIFIWIDGET_H
 #define WIFIWIDGET_H
-#include "widget.h"
+
+#include <widget.h>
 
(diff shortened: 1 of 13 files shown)


ericthelin (approved):

learn-more:
I would suggest to not combine multiple changes into one PR,
because if one of the changes is not acceptable (yet) then it holds back the entire PR.

RobThree:
Yeah, I know. I usually don't. This just happened before I realized 😅

Will do next time.

Thanks!

PR #71, opened 2024-10-22
Support longpress

flattermann:
I added support for long pressing the buttons to add more interaction with the widgets.
I removed the released() function, because it does not really make sense anymore (when you release the button, the state will either be PRESSED or LONGPRESSED).

I implemented it using a new method 
void Widget.changeModeLongpress()

Alternatively, we could also add a bool parameter to changeMode():
void Widget.changeMode(bool longPressed)

Not sure which solution is better...


flattermann:
We could also change
void Widget.changeMode()
to e.g.
void Widget.buttonPressed(uint8_t buttonId)
with some constants (MIDDLE_BUTTON, MIDDLE_BUTTON_LONG...).

The same function could then also be used if we assign e.g. the left button to the widgets as well.
And if you add more buttons in a future PCB release, it would still work without adding more functions. :-)

flattermann:
I now changed the logic to
void Widget.buttonPressed(uint8_t buttonId)

firmware/lib/button/Button.h
```diff
@@ -1,27 +1,55 @@
...
+		const static long LONG_PRESS_TIME = BUTTON_LONG_PRESS_TIME;
+#else
+		const static long LONG_PRESS_TIME = 2000;
+#endif
+
 	private:
 		uint8_t  m_pin;
-		uint16_t m_delay;
-		bool     m_state;
+		bool     m_stateLevel;
```
ericthelin, line 45: I do not like the "stateLevel" name it is confusing and not really state. Maybe "pinLevel"

PR #73, opened 2024-10-22
Check for existence of config.h before trying to include it.

learn-more:
When the file is not included, compilation is terminated and an appropriate error is printed.

This also adds the compiler flag `-Wfatal-errors`, that ensures the compiler stops after the first fatal error.
Usually the first error triggers other errors, so this should help people to focus on the actual issue as well.

Example failure:

image


brett-dot-tech:
I had no idea you could create custom error messages like this, I will very likely be doing more of these, thank you

sjoerdverweij (approved):

ericthelin (approved):

brett-dot-tech (commented):
-

PR #77, opened 2024-10-28
Stock Api & styling updates

brett-dot-tech:
Pushing this to both dev and main(not norm to push right to main, but need to do this as a one-off as the previous stock API has stopped working). The main push also has some small changes brought over from dev as well such as some comment capitalization fixes.

API now supports global markets, crypto,  and forex. See default examples for how to use these.
A small thing to be fixed: the current default font does not support the euro or pound symbol for when showing non-NA stocks, so i just use "EUR/GBP" as a placeholder. Once we address fonts this will be an easy thing to swap out.

Also removed some unused wokwi/display driver code in main now that we're all on the same hardware

Code is poorly commented & formatted,. sorry lol


Info-Orbs/include/model/stockDataModel.h
```diff
@@ -28,8 +35,12 @@ class StockDataModel {
 
    private:
     String m_symbol = "";
+    String m_company = "";
+    String m_currency = "";
     float m_currentPrice = 0.0;
     float m_volume = 0.0;
+    float m_HP = 0.0;
+    float m_LP = 0.0;
```
ericthelin, line 43: maybe m_highPrice and m_lowPrice?

Info-Orbs/src/widgets/stockWidget.cpp
```diff
@@ -92,33 +96,81 @@ void StockWidget::getStockData(StockDataModel &stock) {
...
+    }
+    
+    
+
+// Outputs
+
+
+
+
+    display.fillRect(0,70, screenWidth, 49, TFT_WHITE);  // rgb565 colors
```
ericthelin, line 130: extra whitespace should be removed and the formatting on the comment is off

ericthelin (commented):
Generally looks good but the styling changes should be a separate PR

PR #78, opened 2024-10-28
Stock API&Style changes for dev

brett-dot-tech:
Pushing this to both dev and main(not norm to push right to main, but need to do this as a one-off as the previous stock API has stopped working).

API now supports global markets, crypto,  and forex. See default examples for how to use these.

A small thing to be fixed: the current default font does not support the euro or pound symbol for when showing non-NA stocks, so i just use "EUR/GBP" as a placeholder. Once we adress fonts this will be an easy thing to swap out.

Also removed some unused wokwi/display driver code in main now that we're all on the same hardware

Code is poorly commented & formated,. sorry lol


firmware/lib/config/config.h.template
```diff
@@ -7,15 +7,15 @@
...
 #define WEATHER_UNITS_METRIC //Comment this line out(or delete it) if you want imperial units for the weather
 #define FORMAT_24_HOUR false // Toggle 24 hour clock vs 12 hour clock, change between true/false
 #define SHOW_AM_PM_INDICATOR false // AM/PM on the clock if using 12 hour
 #define SHOW_SECOND_TICKS true // Ticking indicator on the centre clock
 #define INVERTED_ORBS false // Set to true if using InfoOrbs upside down. Inverts screens and re-orders screens and buttons.
 //#define WEB_DATA_WIDGET_URL "" // Use this to make your own widgets using an API/Webdata source
 //#define WEB_DATA_STOCK_WIDGET_URL "http://<insert host here>/stocks.php?stocks=SPY,VT,GOOG,TSLA,GME" // Use this as an alternative to the stock ticker widget
-#define WIDGET_CYCLE_DELAY 0 // Automatically cycle widgets every X ms, set to 0 to disable
+//#define WIDGET_CYCLE_DELAY 0 // Automatically cycle widgets every X ms, set to 0 to disable
```
learn-more, line 18: Why is this commented out?

firmware/src/main.cpp
```diff
@@ -69,16 +69,6 @@ void setup() {
   TJpgDec.setSwapBytes(true); // JPEG rendering setup
   TJpgDec.setCallback(tft_output);
 
-#ifdef GC9A01_DRIVER
-  Serial.println("GC9A01 Driver");
-#endif
-#ifdef ILI9341_DRIVER
-  Serial.println("ILI9341 Driver");
-#endif
-#if HARDWARE == WOKWI
```
learn-more, line 78: I'd keep the Wokwi in here, there is a PR that re-adds some of the Wokwi files.

firmware/src/model/stockDataModel.cpp
```diff
@@ -15,6 +15,34 @@ StockDataModel &StockDataModel::setSymbol(String symbol) {
 String StockDataModel::getSymbol() {
     return m_symbol;
 }
+
+
+
+
+StockDataModel &StockDataModel::setCompany(String company) {
+    if (m_company != company) {
+        m_company = String(company);
```
learn-more, line 24: Why the explicit cast to `String`? It is already a `String`

firmware/src/model/stockDataModel.cpp
```diff
@@ -29,22 +57,41 @@ String StockDataModel::getCurrentPrice(int8_t digits) {
...
+        m_changed = true;
+    }
+    return *this;
+}
+
+float StockDataModel::getLP() {
+    return m_LP;
+}
+
+String StockDataModel::getLP(int8_t digits) {
```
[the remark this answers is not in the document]
learn-more: Or remove it since it seems to be not used

firmware/src/widgets/stockWidget.cpp
```diff
@@ -102,27 +106,71 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+
+
+
+    display.fillRect(0,70, screenWidth, 49, TFT_WHITE);  // rgb565 colors
+    display.fillRect(0,111, screenWidth, 15, TFT_LIGHTGREY);  // rgb565 
+    display.setTextSize(1);  // Set text size
+    display.setTextColor(TFT_WHITE);
+    display.drawString("52 Week:", centre, 188, 1);
+    display.setTextDatum(ML_DATUM);
+    display.drawString("H: " + currencySymbol + stock.getHP(), 89, 201, 1);
```
learn-more, line 138: This creates a string (and formats the float) each time the stocks are drawn, is that intentional?

firmware/src/widgets/stockWidget.cpp
```diff
@@ -102,27 +106,71 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+
+    display.drawString(stock.getSymbol(), centre, 93, 1);
+    display.setTextColor(TFT_WHITE);
+
+    display.drawString(currencySymbol + stock.getCurrentPrice(2), centre, 155, 1);
+
 
-    display.drawString(stock.getPercentChange(2) + "%", centre, 147, 1);
+    //display.fillRect(0,0,screenWidth, 68, TFT_RED);  // rgb565 colors
+    //display.setTextSize(3);  // Set text size
```
learn-more, line 172: Debug leftover?

learn-more (changes requested):

firmware/src/widgets/stockWidget.cpp
```diff
@@ -102,27 +106,71 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+    }
+    
+    
+
+// Outputs
+
+
+
+
+    display.fillRect(0,70, screenWidth, 49, TFT_WHITE);  // rgb565 colors
```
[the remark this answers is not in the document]
sjoerdverweij: Two options:
1. The API wants an actual RGB565 value there, in which case, 49 is not a valid value (is it an index?!)
2. The API wants a regular 32-bit RGB(A).
If the rectangle is dark blue, this is RGB and it should just say 0x000031 (and the comment should go); if it is something else, who knows what is going on.

firmware/src/widgets/stockWidget.cpp
```diff
@@ -102,27 +106,71 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
 
     // Calculate center positions
     int screenWidth = display.width();
     int centre = 120;
+    int arrowOffsetX = 0;
+    int arrowOffsetY = -109;
+    String currencySymbol = "$";
+    if (stock.getCurrency() == "EUR")
+    {
+       currencySymbol = "EUR";
```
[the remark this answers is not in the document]
brett-dot-tech: @learn-more, so the idea here was to use the appropriate unicode symbol for the currency in question (ex. €). However the font does not support such symbols so I'm using the "EUR" as a placeholder in this case. 

Once the fonts are resolved this if statement will allow for us to change out to the proper currency symbol.

I've fixed the formatting though, thank you

firmware/lib/config/config.h.template
```diff
@@ -7,15 +7,15 @@
 #define WIFI_PASS "WIFIPASS" // Wifi password
 #define TIMEZONE_API_LOCATION "America/Vancouver" // Use timezone from this list: https://timezonedb.com/time-zones
 #define WEATHER_LOCAION "Victoria, BC" //city/state for the weather
-#define STOCK_TICKER_LIST "SPY,VT,GOOG,TSLA,GME" // Choose your 5 stocks to display on the stock tracker
+#define STOCK_TICKER_LIST "BTC/USD,USD/CAD,XEQT,SPY,APC&country=Germany" //Choose 5 securities to track. You can track forex, crypto (symbol/USD) or stocks from any exchange (if one tickerr is part of multiple exchanges you can add on "&country=Canada" to narrow down to your ticker)
```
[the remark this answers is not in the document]
brett-dot-tech: Fixed, the country is intentional yes, it's there as an example if anyone wants to identify a ticker in a specific market as stated in the comment

firmware/src/widgets/stockWidget.cpp
```diff
@@ -102,27 +106,71 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+    {
+       currencySymbol = "GBP";
+    }
+    
+    
+
+// Outputs
+
+
+
```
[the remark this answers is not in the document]
brett-dot-tech: Lazy. Fixed

firmware/lib/config/config.h.template
brett-dot-tech: fixed

firmware/src/main.cpp
brett-dot-tech: fixed

firmware/src/model/stockDataModel.cpp
brett-dot-tech: Was just re-using code, fixed.

brett-dot-tech: Also fixed for currency

firmware/src/widgets/stockWidget.cpp
brett-dot-tech: Intentional yes. Would this not be needed for it all to be drawn on one single line as needed? If we wanted to draw the strings and floats separately it'd need two call them both separately, making alignment allot harder.

firmware/src/widgets/stockWidget.cpp
brett-dot-tech: yes, removed, thanks.

firmware/src/widgets/stockWidget.cpp
```diff
@@ -108,29 +112,66 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+    // additionally, in some cases(crypto) the ticker wont return a currency value, so it must be extracted from the symbol deffinition 
+    if (stock.getCurrency() == "EUR"){
+       currencySymbol = currencySymbolEUR;
+    } else if (stock.getCurrency() == "GBP"){
+       currencySymbol = currencySymbolGBP;
+    } else if (stock.getSymbol().indexOf("/EUR") != -1){
+       currencySymbol = currencySymbolEUR;
+    } else if (stock.getSymbol().indexOf("/GBP") != -1){
+       currencySymbol = currencySymbolGBP;
+    }
```
learn-more, line 137: Wouldn't it be simpler to do: `currencySymbol = stock.getCurrency()` for now?
Then it works the same for all currencies, and it isn't only 'fixed' for EUR and GBP.

firmware/src/widgets/stockWidget.cpp
```diff
@@ -108,29 +112,66 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+// Outputs
+    display.fillRect(0,70, screenWidth, 49, TFT_WHITE); 
+    display.fillRect(0,111, screenWidth, 15, TFT_LIGHTGREY);  
+    display.setTextSize(1);  
+    display.setTextColor(TFT_WHITE);
+    display.drawString("52 Week:", centre, 188, 1);
```
learn-more, line 143: Nit: Shouldn't this be `52 Weeks:` ?
(Note that I am not a native English speaker, so I might be wrong here)

firmware/src/widgets/stockWidget.cpp
```diff
@@ -108,29 +112,66 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
...
+    display.setTextDatum(ML_DATUM);
+    display.drawString("H: " + currencySymbol + stock.getHighPrice(), 89, 201, 1);
+    display.drawString("L: " + currencySymbol + stock.getLowPrice(), 89, 211, 1);
+    display.setTextDatum(CC_DATUM);
+    display.setTextColor(TFT_BLACK);
+    display.drawString(stock.getCompany(), centre, 118, 1);
+    display.setTextSize(4); 
     if (stock.getPercentChange() < 0.0) {
```
learn-more, line 151: Is this `getPercentChange` more relevant than a trend?
(I am unsure if this api has a trend)

learn-more (commented):
Some more comments, most are about code style.

firmware/src/main.cpp
```diff
@@ -100,15 +100,12 @@ void setup() {
...
+  #endif
+  #if HARDWARE == WOKWI
   Serial.println("Wokwi Build");
-#endif
+  #endif
```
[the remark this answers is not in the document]
brett-dot-tech: Fixed, copy/paste error

firmware/include/model/stockDataModel.h
```diff
@@ -10,26 +10,39 @@ class StockDataModel {
     StockDataModel();
     StockDataModel &setSymbol(String symbol);
     String getSymbol();
+    StockDataModel &setTicker(String ticker);
```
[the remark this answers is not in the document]
brett-dot-tech: This was just me re-using code from the previous widget/other widgets as inspiration, I'm not sure if its needed, but its used in every single other data model, is there any reason _not_ to have it?

firmware/src/widgets/stockWidget.cpp
brett-dot-tech: Yes, but I will be fixing this very soon here, so I will be leaving it for now.

firmware/src/widgets/stockWidget.cpp
brett-dot-tech: Valid Nit, and you are correct technically.
However in the world of stocks the colloquialism/widely accepted term is as written (also applies to 3 month low/high etc)

Classic english and its weird edge cases

firmware/src/widgets/stockWidget.cpp
brett-dot-tech: Its standard in the stock market for performance to be tracked day/day, trend would be possible, but thats just not how most folks that follow stocks closely expect the data to be represented

firmware/src/widgets/stockWidget.cpp
ericthelin: I believe the reason is that the full version would be something like "52 week high". Which is the highest value from the last 52 weeks. So you are still only referencing one value therefore it is singular.

firmware/src/widgets/stockWidget.cpp
ericthelin: agreed, a trend is a separate thing. Percent change is to help you see that a $1.00 stock going up $0.50 is a much bigger deal than a $1000 stock going up $0.50. So it is easy to show. But to show a trend you really need a graph or something that shows more data points and a time frame over which you are looking at the trend. If someone was really into it we could do a stock view that rotated between stocks with one stock data filling all 5 orbs.

firmware/lib/config/config.h.template
```diff
@@ -9,7 +9,7 @@
 #define WEATHER_LOCATION "Victoria, BC" // City/state for the weather
 #define WEATHER_SCREEN_MODE Light // Can be either Light or Dark 
 #define WEATHER_UNITS_METRIC // Comment this line out (or delete it) if you want imperial units for the weather
-#define STOCK_TICKER_LIST "SPY,VT,GOOG,TSLA,GME" // Choose your 5 stocks to display on the stock tracker
+#define STOCK_TICKER_LIST "BTC/USD,USD/CAD,XEQT,SPY,APC&country=Germany" //Choose 5 securities to track. You can track forex, crypto (symbol/USD) or stocks from any exchange (if one ticker is part of multiple exchanges you can add on "&country=Canada" to narrow down to your ticker)
```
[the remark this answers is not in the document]
ericthelin: I would rather see the country separated into a separate variable from the stock symbols as it really is a different thing. Also the url notation is rather specific to this api

firmware/src/widgets/stockWidget.cpp
ericthelin: All of this logic should be in the data model.

ericthelin (approved):

PR #79, opened 2024-10-28
Show time weather refresh and language support for en,de,fr (month and weekday)

HugoBalder:


HugoBalder:
I will change the branch in my first pull request and hope that works.

PR #84, opened 2024-10-28
Fix jaggies on weather widget

sjoerdverweij:


.vscode/launch.json
```diff
@@ -0,0 +1,44 @@
+// AUTOMATICALLY GENERATED FILE. PLEASE DO NOT MODIFY IT MANUALLY
```
learn-more, line 1: Please remove the files from the `.vscode` folder from your PR

firmware/include/widgets/weatherWidget.h
```diff
@@ -11,17 +11,18 @@
 
 #include "model/weatherDataModel.h"
 
-class WeatherWidget : public Widget {
-   public:
-    WeatherWidget(ScreenManager& manager);
+class WeatherWidget : public Widget
```
learn-more, line 14: I would suggest to refrain from formatting changes, because this will cause merge conflicts for everyone else working on the same files.

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -235,64 +275,77 @@ void WeatherWidget::weatherText(int displayIndex, int16_t b, int16_t t) {
...
 
     display.setTextDatum(MC_DATUM);
     display.fillScreen(TFT_WHITE);
-    display.setTextSize(2);
+    display.setTextSize(1);
 
-    display.fillRect(78, 0, 3, 240, TFT_BLACK);
-    display.fillRect(157, 0, 3, 240, TFT_BLACK);
+    // display.fillRect(78, 0, 3, 240, TFT_BLACK);
+    // display.fillRect(157, 0, 3, 240, TFT_BLACK);
```
learn-more, line 308: debug leftovers?

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -235,64 +275,77 @@ void WeatherWidget::weatherText(int displayIndex, int16_t b, int16_t t) {
...
+            drawDegrees(temperature, x, 122, 6, 1, 7, 4, TFT_BLACK, TFT_WHITE);
         }
 
-        String weekUpdate = dayStr(weekday(m_time->getUnixEpoch() + (86400 * (i + 1))));
-        weekUpdate.remove(3);
-        weekUpdate.toUpperCase();
-        display.drawString(weekUpdate, xOffset, 150, 2);
+        String shortDayName = dayStr(weekday(m_time->getUnixEpoch() + (86400 * (i + 1))));
+        shortDayName.remove(3);
+        // weekUpdate.toUpperCase();
```
learn-more, line 342: Debug leftover?

learn-more (changes requested):
Hello,

Thanks for your contribution!

To make it easier to review, and to not cause a lot of merge conflicts I would suggest to remove the formatting changes from this PR, and only commit the changes that are needed for this feature.

.vscode/launch.json
sjoerdverweij: I don't want to have to do that every time -- is it OK if I remove it, commit, re-add it, then add it to a .gitignore?

PR #86, opened 2024-10-29
Weather widget overhaul (dynamic dark mode, aliasing)

sjoerdverweij:


sjoerdverweij:
https://trello.com/c/J9hgYbu9/25-add-new-themes-styling-fonts-to-existing-widgets

flattermann:
For me, it looks like this:
image

I would like to propose the following changes:

1. Second screen: Center content vertically, make city name a bit smaller and over two lines if necessary, increase vertical space between lines
2. Third screen: Make text bigger (it's really tiny at the moment)
3. Fifth screen: It shows "partly cloudy" overlayed, that looks like a bug
4. I think "High/Low/Cloudy" with a capital character looks better, but that's subjective...

flattermann:
Also, when switching to the ClockWidget, the display is sometimes offset to the right:
image

Strange enough, this only happens sometimes.
Not sure how this PR might be causing this, but I never had this before using this code...

brett-dot-tech:
I really love all these changes (boot screen also made me chuckle, bug jump scare there lol).

I will however echo most of the tweaks to be made by @flattermann, with some modifications:

1) Dont worry about the centre screen vert centring, it will be finicky to do we can worry about it after.
2) **I'd rather just remove the text below the image, for the 3rd screen** I agree it's too small as is and I think it would look better without. 
3) Also seeing this overlay of smaller text on the final screen
4) You can leave high/low capitalization, we can vote on it later, easy change later down the road if we decide.
5) Also experiencing the number offset as well for the clock, this is a big one, i'd imagine this has something to do with adjusting the text datems, as i've had something similar happen in the past.
6) It appears the city name text is still jaggy- was this intentioanl?

sjoerdverweij:
Removed the small text from the 3rd screen (that also fixes the wacky text on the 5th screen, I left in some debug code I created when testing why icons wouldn't display).

The clock centering issue happened to me occasionally from the jump, but it looks like @flattermann fixed that already.

Re-capitalized "High", "Low", and "Highs/Lows". Three people have mentioned it now, I bow to the majority :-)

I added a single pixel between the forecast lines. Note that there can be 4 lines, it's super-tight.

The place name not fitting I'll have to think about. Like I said, there can be 4 lines of forecast. Ideally I'd like to "autosize" the city name (I already conditionally size the weekday name and the degrees in the 3-day forecast so they fit in all situations); the problem is that I can't go off text length ("Elk Grove" fits, "Dortmund" does not). I'll have to see if there is any way to measure proportionally rendered text.

Short weekdays and city name are still mildly jagged because the larger fonts don't have letters. They've changed from scaling a tiny font 3-4 times to scaling the largest one with letters 2 times. That's as good as it gets for now until someone gets down and dirty with the fonts (and someone should, I HATE this crufty Arial ripoff).

flattermann:
This looks very good. Thank you!

Do you also have new icons for Light mode?

brett-dot-tech:
Personally I dont actually like the more 3d icons on the dark theme, they look out of place on the otherwise less polished device. That being said, this is entirely a preference conversation, so we'll merge it as is as look to change them later in whatever direction we decide (whether thats aligning the light theme to have the dark style icons, or finding more flat icons for the dark theme

brett-dot-tech:
Scratch that, I'm still seeing the weird Datum offset carrying over, I can take a look at this myself later tonight but its certainly still happening(pictured) 

Additionally, the button behaviour seems to be super wonky, this likely does not have to do with this PR though, i'll follow up in discord.
Screenshot 2024-11-02 at 1 53 59 PM

flattermann:
@sjoerdverweij:
FYI: Regarding auto sizing the font for the city name to fit horizontally, there is a promising function:

int16_t TFT_eSPI::textWidth(const char *string, uint8_t font) // Returns pixel width of string in specified font

sjoerdverweij:
Bulk reply:

@flattermann yes, I have the same icons with white background (mutsaers_makes submitted both sets). Not going to put too much effort into fixing their compression right now since @brett-dot-tech doesn't seem to like them.

@brett-dot-tech if you don't like the more complicated weather icon look, I think we should ask mutsaers_makes to make a proper flat set for dark mode. With a bit of effort, we could make it switchable at compilation; with a skosh more, we could make it switchable at runtime. My vote would be to keep it a compilation switch, either someone likes the fancy ones, or they like the flat ones. I am actually partial to flat ones but these new ones look SO DAMNED GOOD.

Yes, I am volunteering to do the work to make this switchable if we get a proper flat dark set.

@brett-dot-tech The newly wonky button behavior, I think, is part of the long-press stuff. For example, as it stands, the weather widget switches between high and low on the LEFT button. I don't think that was intended. It should be very easy to fix; if I do any work today, I'll roll that change in. One thing I want to do is test the long-press changes by changing the weather widget to have a short press switch themes, and a long press switch high/low.

@brett-dot-tech The stock ticker being out of alignment is not a function of other things screwing it up, it's merely the stock ticker not properly setting alignment when needed. I had this happen to me with code from before I contributed anything. It is intermittent (go through the widgets one more time and it'll be gone). The fix is going to be the same as for the clock widget, just properly setting the datum before drawing text.

I will also. again, note that from that snapshot, that it looks like that the stocker ticker has the same problems that the weather widget had -- scaling up an 8x8 font beyond all reason. It could do with the same cleanup I did for weather. Unless you really like the pixellated look, which for stocks, I would totally understand.

@flattermann thank you for unearthing that API call. Yes, that would totally fix things. I could properly auto-size just about everything on the weather widget. At this point, though, I don't want to put it into the weather widget; this is universally useful functionality. I want to lift a lot of functionality from the individual widgets up into either the widget class, or create new Screens and Screen classes that take care of all the drawing. That would be a pretty brutal refactor though, and I'm not going to spend the time until 86 gets in.

flattermann:
@sjoerdverweij:
Just FYI: The button behaviour is already fixed in upstream dev. If you rebase to the current dev branch, everything should work.

For the auto font size, I would also recommend to do it in the future, after this PR is merged.

I would not recommend putting work in the current stock ticker, because @brett-dot-tech is already in the process of reworking it (#78)

ericthelin (approved):

brett-dot-tech:
Thanks for clarifications on all fronts @sjoerdverweij.

Heard on all fronts on the weather widget side. While we could make them switchable on compilation, that opens up a very deep rabit-hole of things being "overly configurable" to where everything will get lost in too many settings/things to configure. Let me put some thought into that as its more of a product opposed to a tech decision. In regards to on compilation vs on run time, i'd bias towards on run time. The only reason there being I'd eventually like to move to rending all images/icons from the web (or at least any images/icons that are non-default). This allows us all sorts of liberties in the future, but the main being flash storage not being occupied by a bunch of local images

Understood on the positioning. I didn't fully understand what was happening before but now I do, this is def just a product of the other widgets being poorly designed (on me, when this code was first written I was not thinking forward to folks designing other widgets that might use different datums/sizes than I was working on). This is a great place to start a precedent moving forward to ensure all widgets are independent in their styling. Nice one, thanks for this. I'll ensure this in place in my new stock widget (as I'm sure its not now lol)

The font size was just me not then (and still not fully) understanding how fonts work in this library. I just thought "I want it bigger so I'll make the number bigger". I still dont understand how it was fixed, but i'll lean on ya'll to handle font stuff.

I'll need to make a few small tweaks here to ensure this is compatible to the previous merges. Doing this now then will merge (:

PR #87, opened 2024-10-30
New widget for Parqet.com

flattermann:
I've created a widget for the (german) stock portfolio tracker parqet.com.

This widget uses functions of PR #71 therefore it may be a good idea to merge #71 first.

What this widget does:

You define the ID of your (public!) Parqet portfolio in config.h PARQET_PORTFOLIO_ID

If you want to try the widget without setting up a parqet.com account, you can use the ID 66851bee9c80618bd501c2ed. That's a public portfolio I found on the internet.

The widget then fetches all stocks in your portfolio and shows them. By default, it shows a clock on screen 1 and the stocks on screen 2-5. If you have more stocks, it will cycle automatically through them.

It will also show the total change of your whole portfolio and a performance chart of the current timeframe.

The widget supports these button events:

1. Short press on middle button: Manually cycle through stocks
2. Medium press on middle button: Change timeframe (today, 1 week, 1 month...)
3. Long press on middle button: Change timeframe back to today

You can configure the widget in ParqetWidget.h, e.g. show the total value of your portfolio (by default, this is hidden for privacy reasons).

I have to allocate dynamic memory when I fetch the stocks/charts, hopefully without creating a memory leak.
As far as I can see, it works and the free memory is not shrinking over time, but my C++ skills are rather limited...


sjoerdverweij:
I am very much on board with this, but you'll have to get the long press code in first I think.

I think we're in the same boat: PLEASE TAKE THIS BASIC CHECKIN I HAVE SO MUCH MORE WAITING :-)

firmware/lib/button/Button.cpp
```diff
@@ -3,37 +3,43 @@
 
 
 Button::Button(uint8_t pin)
-    : m_pin(pin), m_delay(100), m_ignoreUntil(0),
+    : m_pin(pin), m_lastStateLevelChange(0),
+      m_stateLevel(RELEASED_LEVEL), m_state(BTN_NOTHING),
       m_hasChanged(false) {}
 
 void Button::begin() {
-#if BUTTON_MODE == INPUT_PULLDOWN
```
brammie15, line 10: Why remove this?

platformio.ini
```diff
@@ -37,6 +37,8 @@ lib_deps =
 	bodmer/TJpg_Decoder@^1.1.0
 	paulstoffregen/Time@^1.6.1
 	arduino-libraries/NTPClient @ ^3.2.1
+	bblanchon/StreamUtils@^1.9.0
```
brammie15, line 40: Not the biggest fan of more libraries since we are running tight on space but i guess it's fine

flattermann: Me neither, but Parqet uses chunked responses and as far as I know, HTTPClient does not support this natively. 

Chunked responses will probably also be needed for other webservices in the future, therefore I think it's not too bad to include the library.

brammie15: Fair enough

PR #94, opened 2024-11-02
Add language support for Month/Weekday in Weatherwidget

HugoBalder:
Changed config.template.h, globaltime.cpp and weatherwidget.cpp to use defined language for Month/Weekday in Weatherwidget


firmware/lib/globalTime/globalTime.cpp
```diff
@@ -3,6 +3,21 @@
...
+#ifdef LOC_DE
+   const char LOC_MONTH[12][10] = {"Januar","Februar","März","April","Mai","Juni","Juli","August","September","Oktober","November","Dezember"}; // Define german for month
+   const char LOC_WEEKDAY[7][11] = {"Sonntag","Montag","Dienstag","Mittwoch","Donnerstag","Freitag","Samstag"}; // Define german for weekday
+#endif
+
+#ifdef LOC_FR
+    const char LOC_MONTH[12][10] = {"Janvier", "Février", "Mars", "Avril", "Mai", "Juin", "Juillet", "Août", "Septembre", "Octobre", "Novembre", "Décembre"}; // Define french for month
+    const char LOC_WEEKDAY[7][11] = {"Lundi", "Mardi", "Mercredi", "Jeudi", "Vendredi", "Samedi", "Dimanche"}; // Define french for weekday
+#endif
+
```
flattermann, line 20: I would move the LOC_MONTH/WEEKDAY definitions to globalTime.h.

This keeps the .cpp cleaner.

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -4,6 +4,21 @@
...
+#ifdef LOC_DE
+   const char LOC_MONTH[12][10] = {"Januar","Februar","März","April","Mai","Juni","Juli","August","September","Oktober","November","Dezember"}; // Define german for month
+   const char LOC_WEEKDAY[7][11] = {"Sonntag","Montag","Dienstag","Mittwoch","Donnerstag","Freitag","Samstag"}; // Define german for weekday
+#endif
+
+#ifdef LOC_FR
+    const char LOC_MONTH[12][10] = {"Janvier", "Février", "Mars", "Avril", "Mai", "Juin", "Juillet", "Août", "Septembre", "Octobre", "Novembre", "Décembre"}; // Define french for month
+    const char LOC_WEEKDAY[7][11] = {"Lundi", "Mardi", "Mercredi", "Jeudi", "Vendredi", "Samedi", "Dimanche"}; // Define french for weekday
+#endif
+
```
flattermann, line 21: If you move the LOC_MONTH/WEEKDAY definitions to globalTime.h, you can remove the redundant definitions here.

flattermann (commented):
Looks good, but please move the definitions to globalTime.h.

Did you check if the fonts actually include the special characters for german/french?

flattermann:
It seems that the fonts do not include umlaute and accents. :-(

flattermann:
In the future, it would be nice to include another locale dependent definition like

#LOC_FORMAT_DAYMONTH "%d. %B"

to choose the format of the day/month string, e.g. in german "2. November" (with a dot).

https://en.cppreference.com/w/cpp/chrono/c/strftime

flattermann:
It's actually quite easy if you add

   const String LOC_FORMAT_DAYMONTH = "%d. %B"; // in strftime format

to all your locale definitions (EN without dot, DE/FR with dot) and add a new function in globalTime.cpp:

```
String GlobalTime::getDayAndMonth() {
#ifdef WEATHER_UNITS_METRIC
    String retVal = LOC_FORMAT_DAYMONTH;
    retVal.replace("%d", String(m_day));
    retVal.replace("%B", m_monthName);
    return retVal;
#else
    return String(m_day) + " " + m_monthName;
#endif
}
```

We can call this in the widgets to get the correct format for the different locales.

flattermann:
@HugoBalder:
I added a PR to the dev branch in your repo.

If you merge it in your dev branch, this PR will automatically be updated (because it references your dev branch).

firmware/lib/config/config.h.template
```diff
@@ -17,6 +17,7 @@
 //#define WEB_DATA_WIDGET_URL "" // Use this to make your own widgets using an API/Webdata source
 //#define WEB_DATA_STOCK_WIDGET_URL "http://<insert host here>/stocks.php?stocks=SPY,VT,GOOG,TSLA,GME" // Use this as an alternative to the stock ticker widget
 #define WIDGET_CYCLE_DELAY 0 // Automatically cycle widgets every X ms, set to 0 to disable
+#define LOC_EN // Language selection for Month and Weekday - values are LOC_EN, LOC_DE, LOC_FR
```
ericthelin, line 20: How about instead of three define vars we create one with a value?
#define LOCALE en

#if LOCALE de
...
#elif LOCALE fr
...
#else
...
#endif

It will be much clearer for people adding or using future languages.

brett-dot-tech: I like this approach @ericthelin, just much more user friendly.  If someone can make this change I'll merge it afterward.

flattermann: To make this work, we need an enum with the languages, right?

enum locales {
    de, en, fr
};

Otherwise de, en, fr are probably not defined...

flattermann: enums are not enough.

I think we can only compare it with other definitions, e.g.

```
#define EN 0
#define DE 1
#define FR 2

#define LOCALE FR

#if LOCALE==DE
...
#elif LOCALE==FR
...
#else
...
#endif
```

flattermann: Not sure if "define EN/DE/FR might interfere with anything and we should add a prefix like "LOC_"?

`#define LOC_EN 0`

But this does not look that nice for the user

`#define LOCALE LOC_EN`

flattermann:
@HugoBalder:
There is still the issue that the fonts only include ascii characters.
Can you please adjust the month/weekday names accordingly?

HugoBalder:
Sorry, i don't know what i should adjust ?!?

flattermann:
Please change the month/weeknames so that they only contain ASCII characters (because the fonts can't handle special characters).

E.g. März -> Maerz, Février -> Fevrier

HugoBalder:
Makes sense :) I changed the 'funny' characters to e.g. Maerz and é to e and this ú to u

flattermann (approved):

ericthelin (approved):

PR #97, opened 2024-11-03
fix: handle the clock overflow

sloopalesci:
This PR addresses the handling of time intervals and potential overflow of the millis() function in the clockWidget.cpp file. The changes ensure accurate time calculations even when the millis() function wraps around to zero after reaching its maximum value.

These changes ensure that the timing logic in clockWidget.cpp is robust against the overflow of the millis() function, providing accurate time interval calculations.

### DIFF
--- a/Info-Orbs/include/widgets/clockWidget.h
+++ b/Info-Orbs/include/widgets/clockWidget.h
@@ -23,7 +23,7 @@ class ClockWidget : public Widget {
     int m_timeZoneOffset;
 
     // Delays for setting how often certain screens/functions are refreshed/checked. These include both the frequency which they need to be checked and a varibale to store the last checked value.
-    long m_secondTimer = 2000;  // this time is used to refressh/check the clock every second.
+    long m_secondTimer = 1000;  // this time is used to refressh/check the clock every second.
     long m_secondTimerPrev = 0;
 
     // WiFiUDP m_udp;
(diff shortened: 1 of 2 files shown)


flattermann:
Shouldn't it be enough to define m_secondTimer and m_secondTimerPrev as unsigned long in clockWidget.h?

Do you think, this complex code is really needed?

flattermann:
BTW: At the moment, m_secondTimerPrev is never updated, it's constantly 0. 

This means that

    if (millis() - m_secondTimerPrev < m_secondTimer && !force) {
        return;
    }

should never be executed in my opinion (the condition is always false) and no overflow should occur.

flattermann:
I can see than m_updateTimer in globalTime.h will overflow and cause problems because it's defined as signed int, but this should only cause problems after ~25 days, not 5 days.

brett-dot-tech:
I'll wait for review of @flattermann comments as C++ and variable types are not my strongsuit, so I cant really advise here on best practices;

There is also some conflicts in the merge due to recent PR merges, but very eager to get this merged in and finally solve this mystery. Thanks again @sloopalesci

flattermann:
I've created PR #98 that changes the types of all variables related to millis() to unsigned long.

I'm not sure if that also solves the issue from this PR.

Maybe someone else could take a look, please?

ericthelin (approved):

brett-dot-tech:
Going to close this PR as it has strayed far form the main codebase and I think the issue at hand is still there, but we can explore later on

PR #99, opened 2024-11-03
add web examples for webDataWidget

ericthelin:
Lots of people have asked about example for webDataWidget. So this adds the examples of both php code and json that I have been using.


brett-dot-tech:
Thanks for these, i'll link to these in the readme soon as well when I rewrite it for all the new stuff

PR #103, opened 2024-11-04
Remove .vscode files from git

flattermann:
Especially extension.json is driving me nuts. :-)


learn-more:
It still makes sense to have the extension.json there?
Since this is a valid recommendation.

flattermann:
If I close VS Code and delete the extensions.json, the same file is recreated automatically.
Therefore, I thought that we don't need to have it in git.

brett-dot-tech:
If we're going to remove these(which I'm not sure if we should/shouldn't yet), extensions included, should we not also be adding them to gitignore? Else they'll very likely slip in again after a few PRs

PR #104, opened 2024-11-04
Shows time when weather was last updated

HugoBalder:
Last update time from visual crossing data will be displayed on second screen below City and forecast.

### DIFF
--- a/firmware/include/model/weatherDataModel.h
+++ b/firmware/include/model/weatherDataModel.h
@@ -5,8 +5,9 @@
 #include <iomanip>
 
 #define NaN -1024.0
-class WeatherDataModel {
-   public:
+class WeatherDataModel
+{
+public:
     WeatherDataModel();
     WeatherDataModel &setCityName(String city);
     String getCityName();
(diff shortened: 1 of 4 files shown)


flattermann (approved):
This looks better, thanks.

BTW: There was no need to close the other PR.
If you change the underlying branch (your "dev" in this case), the PR should be updated automatically.

firmware/include/model/weatherDataModel.h
```diff
@@ -41,6 +41,9 @@ class WeatherDataModel {
     float getDayLow(int num);
     String getDayLow(int8_t num, int8_t digits);
 
+    WeatherDataModel &setLastUpdate(String lastupdate);
```
learn-more, line 44: Why does this return itself?

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -270,6 +271,12 @@ void WeatherWidget::weatherText(int displayIndex) {
         display.drawString(messageArr[i], centre, y);
         y += 22;
     }
+
+    String lastupdate = model.getLastUpdate();
+    display.setTextSize(1);
+    display.setTextColor(TFT_RED);
+    lastupdate.remove(5);
```
learn-more, line 278: It might make sense to add a comment here what part of the string is being removed.

firmware/include/model/weatherDataModel.h
HugoBalder: Sorry, i don't understand the question. What's the problem here ? I copied the code from the other defintions and only replaced the variable.

HugoBalder: It's the same code as here, only with setlastupdate and lastupdate as variable:

WeatherDataModel &setCityName(String city);
String getCityName();

brett-dot-tech:
Love this idea!
I'll need to further test this, I wont have time until tonight and was just able to briefly flash it and look at the functionality.

My first thought though is for this to be fully rolled out I think the styling would need to be refined further, as showing just as the time it's not entirely clear what its showing. The red is also a lot, not in a bad way, but it does draw the eye more than anything else on the widget while being just a secondary piece of information. (I understand why you did this though to allow for it to work on both light & dark widget easily).

That said those are mostly items of preference, at the very least we'll need this nested under config.h as something to enable/disable, as some folks may not want this in exchange for the more minimal/less cluttered look

HugoBalder:
Doesn‘t make sense anymore as it needs a rewrite depending which font is used. With TTF it needs other values as with regular font. Also i can‘t follow some of the comments which make no sense at all, especially from learn-more.

PR #105, opened 2024-11-04
globalTime: Update timezone after DST start/end

flattermann:


flattermann:
Unfortunately, we can't just add/subtract 1 hour to the offset without doing another API request, because not every timezone uses a 1 hour DST delta.

I've delayed the update randomly by up to 5 minutes to avoid flooding the timezone API.

firmware/lib/globalTime/globalTime.cpp
```diff
@@ -127,16 +127,25 @@ bool GlobalTime::isPM() {
 
 void GlobalTime::getTimeZoneOffsetFromAPI() {
     HTTPClient http;
-    http.begin(String(TIMEZONE_API_URL) + "?key=" + TIMEZONE_API_KEY + "&format=json&fields=gmtOffset&by=zone&zone=" + String(TIMEZONE_API_LOCATION));
```
ericthelin, line 130: Rather than removing the fields so we get logs of data we should be able to just add the new field we want to the field list

ericthelin (approved):

firmware/lib/globalTime/globalTime.cpp
ericthelin: I would be good to merge this as is but this tweak would be an improvement

flattermann: The API reply is so small even without filtering that it does not really make a difference, but I've readded the "fields" parameter.

sjoerdverweij (approved):

PR #110, opened 2024-11-14
Add WiFiManager

KanyonKris:
Add WiFiManager to setup WiFi (SSID and password).

### DIFF
--- a/firmware/include/core/wifiWidget.h
+++ b/firmware/include/core/wifiWidget.h
@@ -30,6 +30,7 @@ private:
     String m_connectionString{ "" };
     String m_dotsString{ "" };
     String m_ipaddress{ "" };
+    String m_apssid{ "" };
     int m_connectionTimer{ 0 };
     const int m_connectionTimeout{ 10000 };
 
--- a/firmware/include/core/wifiWidget.h
+++ b/firmware/include/core/wifiWidget.h
@@ -22,7 +22,7 @@ private:
 
     bool m_isConnected{ false };
     bool m_connectionFailed{ false };
-
+    bool m_hardCodedWiFi{false};
     bool m_hasDisplayedError{ false };
     bool m_hasDisplayedSuccess{ false };
     bool m_configPortalRunning{ false }; 
--- a/firmware/include/core/wifiWidget.h
+++ b/firmware/include/core/wifiWidget.h
@@ -25,10 +25,11 @@ private:
 
     bool m_hasDisplayedError{ false };
     bool m_hasDisplayedSuccess{ false };
-
+    bool m_configPortalRunning{ false }; 
     
     String m_connectionString{ "" };
     String m_dotsString{ "" };
+    String m_ipaddress{ "" };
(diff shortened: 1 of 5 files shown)


KanyonKris:
@brett-dot-tech this is a new PR for WiFiManager. Tested it on my Orbs.

brett-dot-tech:
Yo this is amazing dude thank you. got it up and running on my end as well and it works great.
Two main things here functionality before merging. (I am holding off on merging any other merges as well for this so you dont run into any conflicts like last time (: )

1. I want to still retain the ability for users to hardcode their wifi should they please. Both for folks who are following the guide online where it instructs to do this, or for folks who for some reason cant get wifi manager to work.
I think the easiest way to do this would be to keep the old wifi creds fields but have them commented out by default, with the respective functionalities for wifimanager & hardcoded creds being ifndef'd and ifdef'd respectively(there may be better ways to do this, cc. @ericthelin  @learn-more). This should allow folks who want to/need to hard code the ability to do so. We'll eventually move to 100% wifimanager, but while we're still leaning on the config file I want to have it in there if possible. 

2. The messaging on the boot screen when no network is found (Configure WiFi by browsing to 192.168.4.1) is a little unclear for most folks, I think instead we should have it be "Connect to the wifi network Info-Orbs_XX on your computer or phone to configure". As in this case the device is not yet connected to any network so the IP wont be of much use to the user. Let me know if Im misunderstanding/missing something here though

Thanks again - appreciate all the work here.

KanyonKris:
I feel you should move away from hardcoded credentials ASAP, but I see your point. I'll change it to what you outlined tomorrow.

You're right, the instructions are wrong. Not much space in one orb, this is what I came up with:
```
Connect
phone or PC
to WiFi network:
Info-Orbs_XX
192.168.4.1
```
PCs will connect WiFi to the Info-Orbs_XX access point, then will need to point a web browser to 192.168.4.1
Most smartphones will browse to the portal automatically once WiFi connects to Info-Orbs_XX

KanyonKris:
@brett-dot-tech changes made and tested.

brett-dot-tech:
re: hardcoding/manual configs. I agree fully, just while we're in this "transitional" period where there is still other items in the codebase that need to be hardcoded, I think it makes sense to keep the overlap. But I agree with the sentiment entirely. Strictly a product/user focused decision there to keep it, technically it does not make sense.

Ah I didnt even consider the non-auto portal on PCs, nice one.
That said, I've been wanting to re-do the startup screen for some time now (going to swap my face out for my logo for example lol). I think we can dedicate two orbs two the how-to for wifi manager as its such a critical part of setup, but I can tackle that later, for now what you have works well.

Other changes also look good, I just need to test them when I get home, but otherwise we'll merge this in. Thanks again for your patience on this.

KanyonKris:
@brett-dot-tech FWIW, I like your face on startup. The "floating head" made me laugh. In fact I was going to see if I could animate, make it spin to make it even more amusing.

PR #111, opened 2024-11-14
Support for TrueType Fonts (TTF)

flattermann:
This PR tracks my "openfontrender" branch with the TTF test code.


brett-dot-tech:
@flattermann, I wrote some thing in discord here, but excited to get this merged. Posting some of it here as well for the sake of organization:

I do however like the traditional 7 segment font opposed to the 14 for the clock, just cleaner, but that seems like an easy change using deg7 vs deg4

The TTF stuff is pretty far out of my wheelhouse of understanding but im down to merge, all I ask is along with the ready to merge PR we ensure that:
a) there is a clear way to still use old fonts if needed for a fast refresh use case (and if you can outline that super briefly)
b) how to manage/use the fonts for anyone looking to use them in their widgets (this is part of a larger better documentation sweep i'll be doing, but this will be a pretty critical item so want to make sure we have some sort of blerb on how they work if someone wants to use them, even if just here in disc for now)

Also looks to be a small conflict in main after the wifimanager merge, but should be a simple one to resolve.

firmware/include/widgets/clockWidget.h
```diff
@@ -4,6 +4,35 @@
...
 
+#ifndef CLOCK_FONT
+#define CLOCK_FONT DSEG7
+#endif
+
+#ifndef CLOCK_FONT_SIZE
+#define CLOCK_FONT_SIZE 200
+#endif
+
+// Some fonts are not really centered, use this to compensate
```
learn-more, line 15: I assume there is something in openfontrender or freetype that has metrics for this?

firmware/include/widgets/clockWidget.h
```diff
@@ -4,6 +4,35 @@
...
+#define CLOCK_OFFSET_X_DIGITS -10
+#endif
+
+#ifndef CLOCK_OFFSET_X_COLON
+#define CLOCK_OFFSET_X_COLON 0
+#endif
+
+// Not all digits in DSEG7/14 are aligned identically, therefore we need to offset them
+#ifndef CLOCK_DIGITS_OFFSET
+#define CLOCK_DIGITS_OFFSET { {0, 0}, {1, -5}, {0, 0}, {1, 0}, {0, -5}, {0, 0}, {0, 0}, {0, -5}, {0, 0}, {0, 0} }
```
learn-more, line 26: I would assume this info is available in the font file as well.

firmware/include/widgets/weatherWidget.h
```diff
@@ -33,7 +33,7 @@ class WeatherWidget : public Widget {
     int getClockStamp();
     void configureColors();
 
-    int drawDegrees(const String& number, int x, int y, uint8_t font, uint8_t size, uint8_t outerRadius, uint8_t innerRadius, int16_t textColor, int16_t backgroundColor);
+    // int drawDegrees(const String& number, int x, int y, uint8_t font, uint8_t size, uint8_t outerRadius, uint8_t innerRadius, int16_t textColor, int16_t backgroundColor);
```
learn-more, line 36: this can be removed?

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -45,6 +45,7 @@ void WeatherWidget::setup() {
 }
 
 void WeatherWidget::draw(bool force) {
+    m_manager.setFont(DEFAULT_FONT);
```
learn-more, line 48: This does not seem like something the widget drawing needs to be responsible for? (setting the default font)

firmware/lib/screenManager/screenManager.cpp
```diff
@@ -26,12 +32,54 @@ ScreenManager::ScreenManager(TFT_eSPI &tft) : m_tft(tft) {
...
+  } else {
+    Serial.printf("Unable to load TTF font %d\n", font);
+  } 
 }
 
 TFT_eSPI &ScreenManager::getDisplay() {
   return m_tft;
 }
 
+OpenFontRender &ScreenManager::getRender() {
```
learn-more, line 79: Does this renderer need to be exposed?

firmware/lib/screenManager/screenManager.h
```diff
@@ -5,29 +5,86 @@
...
 
-    void clearScreen(int screen);
+    // Helper functions
+    unsigned int calculateFitFontSize(uint32_t limit_width, uint32_t limit_height, Layout layout, const char *str);
+
+    // Draw string functions
+    void drawString(const char *text, int x, int y, unsigned int fontSize, Align align, u_int16_t fgColor, uint16_t bgColor, bool applyScale=true);
+    void drawString(const char *text, int x, int y, unsigned int fontSize, Align align, u_int16_t fgColor);
+    void drawString(const char *text, int x, int y, unsigned int fontSize, Align align);
+    void drawString(const char *text, int x, int y);
```
learn-more, line 65: I don't think these (and other drawing functions) belong in the screenManager class?
Probably a `Renderer` or `Graphics` class.

firmware/lib/screenManager/screenManager.h
```diff
@@ -5,29 +5,86 @@
...
+    void drawLine(int32_t xs, int32_t ys, int32_t xe, int32_t ye, uint32_t color);
+    void drawArc(int32_t x, int32_t y, int32_t r, int32_t ir, uint32_t startAngle, uint32_t endAngle, uint32_t fg_color, uint32_t bg_color, bool smoothArc = true);
+    void drawSmoothArc(int32_t x, int32_t y, int32_t r, int32_t ir, uint32_t startAngle, uint32_t endAngle, uint32_t fg_color, uint32_t bg_color, bool roundEnds = false);
+    void fillTriangle(int32_t x1, int32_t y1, int32_t x2, int32_t y2, int32_t x3, int32_t y3, uint32_t color);
 
 private:
     uint8_t m_screen_cs[5] = {SCREEN_1_CS, SCREEN_2_CS, SCREEN_3_CS, SCREEN_4_CS, SCREEN_5_CS};
     TFT_eSPI& m_tft;
+    OpenFontRender m_render;
+    TTF_Font m_curFont = TTF_Font::NONE;
```
learn-more, line 85: It seems weird to have this 'global' current font?
I'd expect something per widget or 'context' or whatever that is called internally.

firmware/src/core/wifiWidget.cpp
```diff
@@ -34,9 +35,13 @@ void WifiWidget::setup() {
...
   
-    // reset settings - wipe stored credentials for testing
+    // Hold right button when connecting to power to reset wifi settings
     // these are stored by the ESP WiFi library
-    //wifimgr.resetSettings();
+    if (digitalRead(BUTTON_RIGHT) == Button::PRESSED_LEVEL) {
+      wifimgr.resetSettings();
+      m_manager.drawCentreString("Wifi Settings reset", ScreenCenterX, ScreenCenterY + lineHeight, fontSize);
+      delay(messageDelay);
+    }
```
learn-more, line 44: Changes like this do not belong in this PR imo.

firmware/src/core/wifiWidget.cpp
```diff
@@ -55,16 +60,16 @@ void WifiWidget::setup() {
...
-      display.drawCentreString("to WiFi network:", ScreenCenterX, ScreenCenterY, 4);
-      display.setTextColor(TFT_SKYBLUE);
-      display.drawCentreString(m_apssid, ScreenCenterX, ScreenCenterY + lineHeight, 4);
-      display.setTextColor(TFT_GREENYELLOW);
-      display.drawCentreString("192.168.4.1", ScreenCenterX, ScreenCenterY + lineHeight * 2, 4);
+      m_manager.drawCentreString("Connect", ScreenCenterX, ScreenCenterY - lineHeight * 2, fontSize);
+      m_manager.drawCentreString("phone or PC", ScreenCenterX, ScreenCenterY - lineHeight, fontSize);
+      m_manager.drawCentreString("to WiFi network:", ScreenCenterX, ScreenCenterY, fontSize);
+      m_manager.setFontColor(TFT_SKYBLUE);
+      m_manager.drawCentreString(m_apssid.c_str(), ScreenCenterX, ScreenCenterY + lineHeight, fontSize);
```
learn-more, line 70: Why is this `c_str()` required now?
That seems inconvenient.

firmware/lib/widget/widgetSet.cpp
```diff
@@ -56,23 +56,22 @@ void WidgetSet::prev() {
 void WidgetSet::switchWidget() {
   m_screenManager->clearAllScreens();
   getCurrent()->setup();
+  uint32_t start = millis();
   getCurrent()->draw(true);
+  uint32_t end = millis();
+  Serial.printf("Drawing of %s took %d ms\n", getCurrent()->getName().c_str(), (end-start));
```
learn-more, line 62: Please note that printing might slow down stuff as well.
I'd suggest to keep statistics per widget, and print those ones every few seconds or so.

firmware/lib/screenManager/screenManager.cpp
```diff
@@ -14,6 +14,12 @@ ScreenManager::ScreenManager(TFT_eSPI &tft) : m_tft(tft) {
   m_tft.setTextDatum(MC_DATUM);
   reset();
 
+  // I'm not sure which cache size is actually good.
+  // Needs testing.
+  m_render.setCacheSize(128, 128, 8192);
+  setFont(DEFAULT_FONT);
+  m_render.setDrawer(m_tft);
```
learn-more, line 21: OpenFontRenderer probably also has implementations for other `Drawers` or whatever they call it,
would it make sense to strip those files from this PR?

firmware/src/core/wifiWidget.cpp
```diff
@@ -96,27 +101,29 @@ void WifiWidget::update(bool force) {
...
         Serial.println("Connected to WiFi");
         m_isConnected = true;
+        delay(messageDelay);
     } else if (m_connectionFailed && !m_hasDisplayedError) {
         m_hasDisplayedError = true;
-        display.fillRect(0, blankRectTop, ScreenWidth, ScreenHeight - blankRectTop, TFT_BLACK);
-        display.drawCentreString(m_connectionString, ScreenCenterX, ScreenCenterY + lineHeight, 4);
+        m_manager.fillRect(0, blankRectTop, ScreenWidth, ScreenHeight - blankRectTop, TFT_BLACK);
+        m_manager.drawCentreString(m_connectionString.c_str(), ScreenCenterX, ScreenCenterY + lineHeight, fontSize);
+        delay(messageDelay);
```
learn-more, line 126: Why are those delays here?
That seems unrelated to drawing.

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -128,31 +129,21 @@ bool WeatherWidget::getWeatherData() {
...
-    const int dateY = 197;
+    const int clockY = 120;
+    const int dayOfWeekY = 190;
+    const int dateY = 50;
 
     m_manager.selectScreen(displayIndex);
+    m_manager.fillScreen(m_backgroundColor);
+    m_manager.setFontColor(m_foregroundColor);
 
-    TFT_eSPI &display = m_manager.getDisplay();
```
learn-more, line 137: In this case I don't think the manager should be used directly for drawing, but for future reference if you want to keep a diff smaller then something like this could be used so that all other lines don't need changing:
```cpp
auto& display = m_manager;
```
That allows to keep the code below doing this kind of stuff intact:
```cpp
    display.fillScreen(m_backgroundColor);
    display.setTextColor(m_foregroundColor);
    display.setTextSize(1);
```

firmware/src/widgets/weatherWidget.cpp
```diff
@@ -251,91 +242,87 @@ void WeatherWidget::weatherText(int displayIndex) {
...
         shortDayName.remove(3);
-        display.drawString(shortDayName, x, 154, 4);
+        m_manager.drawString(shortDayName.c_str(), x, 154, fontSize, Align::MiddleCenter);
     }
 
 }
 
-int WeatherWidget::drawDegrees(const String& number, int x, int y, uint8_t font, uint8_t size, uint8_t outerRadius, uint8_t innerRadius, int16_t textColor, int16_t backgroundColor) {
-    TFT_eSPI &display = m_manager.getDisplay();
+// int WeatherWidget::drawDegrees(const String& number, int x, int y, uint8_t font, uint8_t size, uint8_t outerRadius, uint8_t innerRadius, int16_t textColor, int16_t backgroundColor) {
```
learn-more, line 306: this function can probably be removed?

firmware/src/main.cpp
```diff
@@ -23,7 +23,7 @@
 TFT_eSPI tft = TFT_eSPI();
 
 #ifdef WIDGET_CYCLE_DELAY
-unsigned long m_widgetCycleDelay = WIDGET_CYCLE_DELAY;  // Automatically cycle widgets every X ms, set to 0 to disable
+unsigned long m_widgetCycleDelay = WIDGET_CYCLE_DELAY * 1000;  // Automatically cycle widgets every X seconds, set to 0 to disable
```
learn-more, line 26: Unrelated change

firmware/src/widgets/clockWidget.cpp
```diff
@@ -114,33 +109,58 @@ void ClockWidget::buttonPressed(uint8_t buttonId, ButtonState state) {
...
+DigitOffset ClockWidget::getOffsetForDigit(const String &digit) {
+    if (digit.length() == 0) {
+        return {0, 0};
+    }
+    char c = digit.charAt(0);
+    if (c >= 48 && c <= 57) {
+        return m_digitOffsets[c-48];
+    } else {
+        return {0, 0};
+    }
```
learn-more, line 121: This should probably come from the font.

firmware/src/model/weatherDataModel.cpp
```diff
@@ -53,7 +53,7 @@ float WeatherDataModel::getCurrentTemperature() {
 }
 
 String WeatherDataModel::getCurrentTemperature(int8_t digits) {
-    return Utils::formatFloat(m_currentWeatherDeg, digits);
+    return Utils::formatFloat(m_currentWeatherDeg, digits) + "°";
```
learn-more, line 56: I'd suggest to not hardcode this here, but either make it a member of the class or a define somewhere.

firmware/src/widgets/parqetWidget.cpp
```diff
@@ -210,8 +211,8 @@ void ParqetWidget::updatePortfolio()
                 }
                 else
                 {
-                    String id = holding["_id"].as<String>();
-                    Serial.printf("Invalid type: %s, id: %s\n", type.c_str(), id.c_str());
+                    // String id = holding["_id"].as<String>();
+                    // Serial.printf("Invalid type: %s, id: %s\n", type.c_str(), id.c_str());
```
learn-more, line 215: This seems unrelated

firmware/src/widgets/parqetWidget.cpp
```diff
@@ -340,75 +341,54 @@ void ParqetWidget::updatePortfolioChart()
...
     m_manager.selectScreen(displayIndex);
-    TFT_eSPI &display = m_manager.getDisplay();
-    display.fillScreen(background);
+    m_manager.fillScreen(background);
 }
 
 void ParqetWidget::displayClock(int8_t displayIndex, uint32_t background, uint32_t color, String extra, uint32_t extraColor)
 {
-    Serial.printf("displayClock at screen %d\n", displayIndex);
+    // Serial.printf("displayClock at screen %d\n", displayIndex);
```
learn-more, line 349: You have disabled multiple of these prints,
I am unsure about whether or not they need to be kept, but commented out is not the way to go.

learn-more (changes requested):
There are various comments for unrelated changes,
some small mistakes and possible improvements,
and some more fundamental suggestion.

Please read the comments first before going in to change stuff, because changes for the fundamental changes might completely invalidate other comments.

Also, please do not mistake the directness of the comments with rudeness, in my experience sugar-coating review comments obfuscates the actual comment.

flattermann:
@learn-more Thanks for the review. 
I'll answer above to some of the specific comments, but describe my thoughts about the new ScreenManager features here.

I **really** like the convenience to access the drawing/drawString functions directly in ScreenManager and be able to just do this:

```
m_manager.selectScreen(...);
m_manager.drawString(...);
```

without the need to access an intermediate class.
That's really easy to use and we have all draw related functions in one place (regardless if they use TFT_eSPI or OpenFontRender or whatever). I don't really see an advantage in adding a separate Drawing class to it.

Also, the ScreenManager did not really do anything except initialising the TFT_eSPI, so it seemed like a waste. :-)

firmware/lib/screenManager/screenManager.h
```diff
@@ -5,29 +5,86 @@
...
+// All available TTF fonts
+enum TTF_Font {
+    NONE,
+    ROBOTO_REGULAR,
+    FINAL_FRONTIER,
+    DSEG7,
+    DSEG14
+};
+
+struct TTF_FontScale {
```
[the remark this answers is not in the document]
flattermann: Ok, I'll rename it.

firmware/lib/OpenFontRender/.clang-format
```diff
@@ -0,0 +1 @@
+{ BasedOnStyle: LLVM, UseTab: ForIndentation, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Attach, AllowShortIfStatementsOnASingleLine: false, AllowShortCaseLabelsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All, FixNamespaceComments: false, AlignConsecutiveAssignments: true, AlignAfterOpenBracket: true, AlignTrailingComments: true, IndentPPDirectives: BeforeHash }
```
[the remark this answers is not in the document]
flattermann: I copied the vanilla OpenFontRender folder there.
I'm not sure if it's a good idea to remove files from it (even if they are only format files).
Do they harm us in any way?

firmware/include/ttffonts.h
```diff
@@ -0,0 +1,26 @@
+#ifndef TTFFONTS_H
+#define TTFFONTS_H
+
+#include <Arduino.h>
+
+// These symbols are generated from the files specified in platformio.ini under 'board_build.embed_files'
+// See https://docs.platformio.org/en/latest/platforms/espressif32.html#embedding-binary-data for more info
+
+extern const byte robotoRegular_start[] asm("_binary_fonts_RobotoRegular_ttf_start");
```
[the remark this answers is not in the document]
flattermann: I used the icons.h as a template for this file.
I think it's less cluttered if it's in a separate file, don't you think?

Of course, I can see the advantage of having this in the same place as the TTF_Font enum and the TTF_FontMetric struct.
Maybe move the enum/struct here?

firmware/include/widgets/clockWidget.h
flattermann: Unfortunately, I could not find any other way to show the clock reliably without it.

And I tried a lot of things with getHeight(), calculateBoundingBox() and other stuff.

But feel free to give it a try. I would be **very** happy if we can avoid/reduce the manual compensations.

firmware/include/widgets/clockWidget.h
flattermann: The clock with shadow is especially picky because we first draw a 8 (in bg color) and then the actual number (in fg color).
If the character has even the slightest offset, we are screwed.

And I could not find a way to calculate this offset with the OpenFontRender methods.

But, same as above, I would be glad if someone could figure it out.

firmware/src/widgets/weatherWidget.cpp
flattermann: It's more like the WeatherWidget saying that it wants to use the default font.

The selected font is global at the moment (see other comment), therefore each widget is responsible that it sets the font that it wants to use before calling drawString().

I would prefer having some kind of lifecycle like Android with onStart(), onPause(), onResume(), onStop().

Then we could put .setFont() in onResume().
We can't really put it in setup() because that's called for all widgets during boot. For me, setup() is like onStart() at the moment...

firmware/lib/screenManager/screenManager.cpp
flattermann: I would like to set .getRender() and .getDisplay() to private.
Then all drawing would have to go through ScreenManager.

It did not do it, because I've not implemented delegates for all TFT_eSPI and OpenFontRender functions yet.
But we could set them to private and just add the rest of the delegate functions when we need them.

firmware/lib/screenManager/screenManager.h
flattermann: Well, I see your point, but I also see the convenience this offers.
See also my main comment below.

firmware/lib/screenManager/screenManager.h
flattermann: Yes, the font is global, but you can switch between the fonts anytime.
We can/should only load one font at a time because of memory constraints.

firmware/src/core/wifiWidget.cpp
flattermann: Altough you are technically right, I could not try the WifiWidgets TTFs without it and I just kept it in.
It's such a small change that a separate PR is not really necessary IMHO.

firmware/src/core/wifiWidget.cpp
flattermann: OpenFontRender uses c-Strings.
But I can change the signature to String if that's preferred (or add it as alternative).

firmware/lib/widget/widgetSet.cpp
flattermann: This is only printed once if you change the widget (by pressing a button or by the automatic widget cycling). 
Therefore, I don't think that will have much impact.
But I can remove it if you want to.

firmware/lib/screenManager/screenManager.cpp
flattermann: I don't think there are specific files for the other drawers (it's more like an interface definition calling the same functions like .drawPixel() ).

firmware/src/core/wifiWidget.cpp
flattermann: Without the delays, the screens are immediately cleared and you can't see the message.

firmware/src/widgets/weatherWidget.cpp
flattermann: See comment above.

firmware/src/model/weatherDataModel.cpp
flattermann: Do you mean to define the Degree symbol somewhere else?
What's the point of that? If we want to change the temperature to Kelvin or something?

firmware/lib/globalTime/globalTime.h
```diff
@@ -16,14 +16,17 @@
    const char LOC_MONTH[12][10] = {"Januar","Februar","Maerz","April","Mai","Juni","Juli","August","September","Oktober","November","Dezember"}; // Define german for month
    const char LOC_WEEKDAY[7][11] = {"Sonntag","Montag","Dienstag","Mittwoch","Donnerstag","Freitag","Samstag"}; // Define german for weekday
    const String LOC_FORMAT_DAYMONTH = "%d. %B"; // in strftime format
+   const String LOC_LANG = "de";
```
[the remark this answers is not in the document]
flattermann: I used the WeatherWidget to test special characters (äöüéà...) by adding this to get localized weather texts for some languages. It looked nice, so I kept it in.

firmware/lib/screenManager/screenManager.h
```diff
@@ -5,29 +5,86 @@
 #include <config_helper.h>
 #include <TFT_eSPI.h>
 #include <SPI.h>
+#include <OpenFontRender.h>
+#include "../../include/ttffonts.h"
```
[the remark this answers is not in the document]
flattermann: ScreenManager.setFont() needs the references to the start/end of the fonts.

BTW: Do you know, why 

#include <ttffonts.h>

does not work and I have to use this ugly ".../.../ttffonts.h"?

flattermann:
I have to admit, that this PR includes some minor improvements not related to TTFs.
That's because I used this branch as my main dev branch for quite some time and did not switch branches for these small adjustments.
I don't think that this is a problem and I would like to avoid having to separate these small changes into separate PRs.
But feel free to not merge some of the changes if you prefer that.

firmware/src/widgets/clockWidget.cpp
flattermann: I don't know how, but I'm open to suggestions.

firmware/src/core/wifiWidget.cpp
flattermann: I've changed the functions to use Strings

firmware/include/widgets/weatherWidget.h
flattermann: I removed the commented code

firmware/src/widgets/weatherWidget.cpp
flattermann: I removed the commented code

brett-dot-tech:
@flattermann this looks great; only issue to note is the first orb in the clock widget draws a few pixels offset in 12H mode, causing it to look "sku'd" while booting with 12H by default, then odd overlapping if you switch between 12/24 hour modes using the centre button

PR #124, opened 2024-11-26
Bugfix Weatherwidget always refreshing data when selected

HugoBalder:
Everytime you switch to the weather widget, the data get's reloaded, regardsless of the set refresh time of e.g. 600000 miliseconds which is 10 Minutes. This change fixes this.


brett-dot-tech:
Good catch here @HugoBalder, though I hadnt gone through it yet auditing this was on my list after the auto-widget-switch was enabled in interest of preserving API credits

PR #153, opened 2024-12-03
Fix MQTT widget orb label/text X position

dreed47:
Enables the configuration of the orb label/text X position in the MQTT message.


brett-dot-tech:
@nutk

PR #155, opened 2024-12-04
Switch between Highs/Lows every 10 seconds

HugoBalder:
Normally you would press the middle button to show the Highs or Lows for the next 3 days on the 5th screen. With this change it switches every 10 seconds automatically. The delay is needed, as it would switch several times during 1 second, because the draw routine is called very often from main.cpp.

### DIFF
--- a/firmware/src/widgets/weatherwidget/WeatherWidget.cpp
+++ b/firmware/src/widgets/weatherwidget/WeatherWidget.cpp
@@ -59,6 +59,14 @@ void WeatherWidget::draw(bool force) {
         threeDayWeather(4);
         model.setChangedStatus(false);
     }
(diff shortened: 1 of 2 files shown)


learn-more:
a delay like this is a hack, that blocks all other processing in this thread of the esp.
I would suggest to keep track of the 'already switched' instead of just blocking everything.

brett-dot-tech:
I like this functionally as a concept. However I think it at the very least needs to be behind a config setting.
I personally dont like the "flashing" on an otherwise pretty chill/non disruptive widget.

Even then I feel there may be better ways to do highs/lows than we're doing now. Thoughts on having both?
x/x for example? or  maybe there another common notation common weather apps use?  I built the current layout very hastefully and its just kind of stuck so far, very open to re-working it

Issue #148, opened 2024-12-01
Make network calls async


flattermann:
We should try to make the network calls asynchronous.

At the moment, the UI is completely blocked during updates. This is bad for UX, because the buttons don't work and the Orbs are "frozen".

lyricnz:
Also, it would be good to be able to switch from the "loading screens" to the first widget before all the widgets had updated their data (faster bootup time)

brett-dot-tech:
@dreed47 made some huge progress here, but keeping this issue open as the async fix has resulted in some minor UX bugs #171 , and we have yet to roll it out globally to other widgets (weather). @learn-more suggested some sort of globalized function to handle such async calls which I think sounds like a good idea

flattermann:
We just have to make sure that only one HTTP request is running at a time, because a HTTPClient with ArduinoJson parsing will easily use 70-100kB RAM and at the moment, the Orbs only have ~130kB free RAM when idle.
So if we switch all widgets to async/tasks, this will become an issue at boot.

dreed47:
Maybe we could use a Task semaphore to only allow one HTTP request at a time.  It would require us moving all the HTTP requests to use Tasks.  Or possible create a HTTP helper class or method the we use instead that wraps the Task and semaphore stuff and maybe JSON filtering too?

lyricnz:
This is essentially #222 now, right?

dreed47:
yes I think so

flattermann:
Implemented for WeatherWidget, StockWidget and ParqetWidget now.

PR #160, opened 2024-12-06
Fix UI blocking issue for stock widget by moving http calls to their …

dreed47:
Per issue  #148.  This code change moves the stock widget HTTP calls to their own task.  This frees the UI to accept left/right button clicks.


firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -46,11 +56,8 @@ void StockWidget::draw(bool force) {
 void StockWidget::update(bool force) {
     if (force || m_stockDelayPrev == 0 || (millis() - m_stockDelayPrev) >= m_stockDelay) {
         setBusy(true);
-        for (int8_t i = 0; i < m_stockCount; i++) {
-            getStockData(m_stocks[i]);
-        }
-        setBusy(false);
-        m_stockDelayPrev = millis();
+        // Create a task to handle all stock updates
+        xTaskCreate(taskGetStockData, "StockDataTask", 8192, this, 1, NULL);
```
learn-more, line 60: How do you ensure that this task is not created again while it is running?

dreed47: good point...maybe need to make use of the TaskHandle_t in FreeRTOS to check first if its running.  I'll try some testing for that

firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -6,7 +6,23 @@
 
 #include <iomanip>
 
-StockWidget::StockWidget(ScreenManager &manager) : Widget(manager) {
+void StockWidget::taskGetStockData(void *pvParameters) {
```
lyricnz, line 9: Normally the constructor should be the first in the file.

PR #165, opened 2024-12-09
Fix UI blocking issue for stock widget 

dreed47:
Per issue https://github.com/brettdottech/info-orbs/issues/148. This code change moves the stock widget HTTP calls to their own task. This frees the UI to accept left/right button clicks.


brett-dot-tech:
@dreed47 can you please do this in the dev branch instead? trying to reserve main for major version rollouts

PR #169, opened 2024-12-10
Fix UI blocking issue for stock widget

dreed47:
Per issue https://github.com/brettdottech/info-orbs/issues/148. This code change moves the stock widget HTTP calls to their own task. This frees the UI to accept left/right button clicks.

### DIFF
--- a/firmware/src/widgets/stockwidget/StockWidget.cpp
+++ b/firmware/src/widgets/stockwidget/StockWidget.cpp
@@ -6,7 +6,7 @@
 
 #include <iomanip>
 
-StockWidget::StockWidget(ScreenManager &manager) : Widget(manager) {
+StockWidget::StockWidget(ScreenManager &manager) : Widget(manager), m_taskHandle(NULL) {
 #ifdef STOCK_TICKER_LIST
     char stockList[strlen(STOCK_TICKER_LIST) + 1];
     strcpy(stockList, STOCK_TICKER_LIST);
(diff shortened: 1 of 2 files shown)


firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -64,7 +71,7 @@ void StockWidget::buttonPressed(uint8_t buttonId, ButtonState state) {
 }
 
 void StockWidget::getStockData(StockDataModel &stock) {
-    String httpRequestAddress = "https://api.twelvedata.com/quote?apikey=e03fc53524454ab8b65d91b23c669cc5&symbol=" + stock.getSymbol();
+    String httpRequestAddress = "http://api.twelvedata.com/quote?apikey=e03fc53524454ab8b65d91b23c669cc5&symbol=" + stock.getSymbol();
```
learn-more, line 74: Why is this changed from https to http?

firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -101,6 +108,21 @@ void StockWidget::getStockData(StockDataModel &stock) {
     http.end();
 }
 
+void StockWidget::taskGetStockData(void *pvParameters) {
+    StockWidget *widget = static_cast<StockWidget*>(pvParameters);
+    for (int8_t i = 0; i < widget->m_stockCount; i++) {
+        widget->getStockData(widget->m_stocks[i]);
+    }
+    
+    UBaseType_t highWater = uxTaskGetStackHighWaterMark(NULL);
```
learn-more, line 117: This is for debugging?
Should this remain here?

learn-more (approved):
Overall lgtm,
we might want to create a more generic framework for handling http requests in a task later on.

brett-dot-tech:
Just tested this, works very well.

The only "issue" that has resulted from a fallback of this is that when booting, the device no long has to wait while it "loads" the stock data. While this is technically a good thing, it results in when the user first boots the device and if they switch over to the stock widget, it may be blank as the API may still be waiting.

I am testing this on very slow wifi at the airport mind you so this issue is exaggerated here, but some food for thought

PR #170, opened 2024-12-10
weathertext truncates data

HugoBalder:
With the current settings for character length 18 the weather text does not fit on the screen and is truncated. Instead i use 24 characters per line and also the start line is on 115 instead of 125.

before
after


brett-dot-tech:
I like this, I think moving up the text  just looks objectively better as well. thanks!

Issue #174, opened 2024-12-10
Stock Widget should make only one API call for all stocks


dreed47:
As the twelvedata API supports requesting multiple stock quotes in one call the Stock Widget should make only one API call for all stocks.

PR #177, opened 2024-12-11
Changes to pull all stocks in one request.

dreed47:
per issue #174.  This change pulls all 5 stocks in one request to save time and resources

### DIFF
--- a/firmware/src/widgets/stockwidget/StockWidget.cpp
+++ b/firmware/src/widgets/stockwidget/StockWidget.cpp
@@ -70,12 +70,14 @@ void StockWidget::buttonPressed(uint8_t buttonId, ButtonState state) {
         changeMode();
 }
 
-void StockWidget::getStockData(StockDataModel &stock) {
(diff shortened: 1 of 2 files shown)


brett-dot-tech:
@dreed47 sorry for the delay here, there is one issue here in doing it this way, and its more of a product of my previously jank solution to allowing users to declare country in their stock list. See the last ticker in the default config, which has &country=Germany appended.

Because each ticker was pulled individually before, having the country flag tagged onto each request was not an issue, but it now seems that by setting the country flag on even the last ticker, like in the example, it localized every stock in the request; This breaks 3 tickers in the example. 

I'm not sure of a clean way around this and I'd need to look into the twelvedata API, but we need to retain the ability for users to localize their stocks to a given marker on a stock by stock basis

dreed47:
Yeah, setting the &country parameter affects the whole http request, that's the way their API works and they don't provide a method to request multiple stocks AND set a country different for each stock all in the same API request.  One approach is to look for any &country string in any of the stocks and if we find it we fallback to single request for each stock, otherwise get them all at once?

PR #180, opened 2024-12-12
Fix UI blocking issue for weather widget

dreed47:
Per issue https://github.com/brettdottech/info-orbs/issues/148. This code change moves the weather widget HTTP calls to their own task.


brett-dot-tech:
@dreed47 some small conflicts here as a result of the JSON filtering merged previously. Sorry about that, should be an easy resolve

dreed47:
I mistaken backed out a couple changes...ill resubmit this PR

PR #182, opened 2024-12-12
feat: use rotary encoder option

angelius1:
I use a rotary encoder for navigation. You may find the code useful.


brett-dot-tech:
@angelius1 this is a really great feature, and I may one day implement it in the hardware, but for now I will leave it out of the main code base as it wont be of use to the majority of  users with a kit, but I've made a note to link your PR if anyone asks about encoders, thanks (:

PR #183, opened 2024-12-14
Special characters for French corrected for Months

HugoBalder:
Now as we have True Type Fonts, the special characters for February and August in French are now correct -> Février and Août. Before it was Fevrier and Aout.


brett-dot-tech:
Blindly merging this due to my lack of French-speaking-ness, but I trust @HugoBalder has it right (;

ebretteville:
I'm French and it's correct. Thanks for us

PR #185, opened 2024-12-15
Multi custom clocks (and allow dynamic file embedding)

flattermann:
This PR adds the ability to have multiple custom clocks (up to 10).

Also we are able to switch embedding of files on and off just by using a `#define` in `config.h` (or by passing `-D XXX` in `platformio.ini`).
This was not possible before, because PlatformIO does not allow you to dynamically select files to embed based on compile flags. You need to define the files in `platformio.ini` -> board_build.embed_files.
I've added a pre-action build script (in Python) that will make this possible (`scripts/embed_files.py`).
It can be configured at the top of the script.

The `embed_map` is a map of compile flags to directories/files that should be included.

The easiest usage is to say "include files dir1/test1.jpg and dir1/test2.jpg when compile flag TEST is present".
You would do this by setting

`embed_map = { 
  "TEST": 
    [ "dir1/" , 
      [ "test1.jpg", "test2.jpg" ]
    ]
}`

You can also set a condition, like "if TEST > 0 include test1.jpg", "if TEST > 1 also include test2.jpg":

`embed_map = {
  "TEST > 0": 
    [ "dir1/" , 
      [ "test1.jpg" ] 
    ], 
  "TEST > 1": 
    [ "dir1/", 
      [ "test2.jpg" ] 
    ] 
}`

This makes this solution very flexible.

I added code for `#define USE_CLOCK_CUSTOM 0 .. 10` (to embed the respective files based on the number of custom clocks specified) and for the Nixie style `#define USE_CLOCK_NIXIE NIXIE_HOLES` or `#define USE_CLOCK_NIXIE NIXIE_NOHOLES`.

**It's based on the WebPortal branch, so please merge after #159.**


flattermann:
I think we should skip this PR and jump straight to the LittleFS branch, because that makes some stuff in this PR obsolete.

flattermann:
Superseded by #190

PR #191, opened 2024-12-21
Make Weather Feed pluggable

dreed47:
This change refactors out the getWeather method so that people can swap out the feed source if they choose to not use Visual Crossing for their weather data.  It also adds a feed for Tempest weather stations which are a popular personal weather stations.  With this change it would make it relatively easy to add other weather feeds like OpenWeatherMap, WeatherAPI, Pirate Weather, Open-Meteo, or in the US the National Weather Service.


dreed47:
Closing this PR as it needs reworked

PR #192, opened 2024-12-22
Refine position of seconds indicator in Clock Widget

ScottWell1:
Not sure I am doing this right.  I am not a C++ guy, and don't understand GitHub.  But the seconds indicator on the clock has been bugging me.  It is not "centered" over the second; rather the arc starts at the correct degree for the second and then is padded 6 degrees forward. Very noticeable, especially at 0, 15, 30, 45 seconds.

This PR (which may be coded horribly, I don't do c++) fixes the location of the clock seconds indicator to be centered (rather than begun) on the correct angle, 3 degrees left and 3 degrees right of the precise second, for the seconds/360 value.


ScottWell1:
Only the changes to ClockWidget.cpp were intended.  I'm bad at Github, lol.  I think the later commits reverted everything except the ClockWidget changes.

flattermann (approved):
Looks good. Thanks!

PR #195, opened 2024-12-23
Switch between Highs and Lows in Weatherwidget

HugoBalder:
Every 10 seconds the screen with the 3 day forecast switches between the Highs and Lows values. So you don't have to press the button to switch. The time can be set in the weatherwidget.h.

### DIFF
--- a/firmware/src/widgets/weatherwidget/WeatherWidget.h
+++ b/firmware/src/widgets/weatherwidget/WeatherWidget.h
@@ -69,6 +69,10 @@ private:
     const int MODE_HIGHS = 0;
     const int MODE_LOWS = 1;
 
+#ifndef HIGH_LOW_INTERVAL
+  #define HIGH_LOW_INTERVAL 0
+#endif
+
     int m_switchinterval = HIGH_LOW_INTERVAL;
     unsigned long m_prevMillisSwitch = 0;
 };
--- a/firmware/src/widgets/weatherwidget/WeatherWidget.h
+++ b/firmware/src/widgets/weatherwidget/WeatherWidget.h
@@ -68,5 +68,9 @@ private:
 
     const int MODE_HIGHS = 0;
     const int MODE_LOWS = 1;
+
+    const long switchinterval = 10000;
+    unsigned long currentSwitchMillis;
+    unsigned long prevMillisSwitch = 0;
 };
 #endif // WEATHERWIDGET_H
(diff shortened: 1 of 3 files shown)


brett-dot-tech:
Hey! I recall someone merging something similar to this before. 

I think the flashing of the high/lows might be too much for some folks. I also want to make sure we are keeping all user configureable variables in config.h.

I'd ask that we ensure this is disabled by default, as well as the enable/time config be placed in config.h.

Thanks (:

HugoBalder:
Hi Brett, i could add config.h values and use them in the code as next step. Will upload the changes soon.

HugoBalder:
Hi Brett, now it can be configured and default is OFF. Hope this is ok now for merge :)

flattermann:
Can you maybe make it configurable via WebPortal?
i.e. change the variable type of switchinterval to int, move the *1000 to the comparision in the if clause and add something like
`config.addConfigInt("WeatherWidget", "weatherCycleHL", &switchinterval, "Switch between Highs and Lows every X seconds, set to 0 to disable", true);`
to the WeatherWidget constructor.

flattermann:
BTW: I really like this feature and would prefer to **enable** it by default, but that's for @brett-dot-tech to decide.

HugoBalder:
Hi Flattermann, i implemented your suggestion, added the configuration option and tested it. One thing i always notice, when you save your changes. The ESP reboots too fast and you are left with an uncompleted 'Saved...' Screen in the Browser. Seems 1 second delay is too short.

flattermann (changes requested):

firmware/src/widgets/weatherwidget/WeatherWidget.h
```diff
@@ -68,5 +68,9 @@ class WeatherWidget : public Widget {
 
     const int MODE_HIGHS = 0;
     const int MODE_LOWS = 1;
+
+    int switchinterval = HIGH_LOW_INTERVAL;
```
flattermann, line 72: Maybe add
```
#ifndef HIGH_LOW_INTERVAL
  #define HIGH_LOW_INTERVAL 0
#endif
```
on top of the file to set a default value in case it's not defined in config.h?

brett-dot-tech:
> BTW: I really like this feature and would prefer to **enable** it by default, but that's for @brett-dot-tech to decide.

I will put this to a vote then, @flattermann thanks for the feedback on this to @HugoBalder , i'll merge this once resolved, standby for results from the community on whether its enabled/disabled by default (:

HugoBalder:
I changed the variable names, fixed the code and the default is set to OFF (was already in the first upload of the config.template.h). Saw the poll in discord and the majority doesn't want it to be ON by default.

flattermann (approved):
Looks good, thanks!

PR #200, opened 2024-12-27
Drawscript- simple line based script to generate widget drawing code server-side

dejaloomer:
This patch adds a simple scripting system to expose the ScreenManager's drawing functions. This enables drawing widgets from a host/server.
The patch hooks this ability into mqtt, though it could be used in any widget that can receive strings.
Also fixed the mqtt connection retry so it doesn't block in the widget.


brett-dot-tech:
I'll look a bit deeper into this in a sec but have you looked at @ericthelin's web data widget? Based on your comment it sounds like its very similar  in functionality to this

dejaloomer:
yeah, I didn't realize what webdata was until after I had done this. :^/
However, I think this script is significantly easier to generate and more compact than the json in the webdata. Since it maps directly to ScreenManager drawing calls it also makes for easy prototyping of widget draw code.
Since it just parses a string, it could also be integrated into WebData and provide an additional mechanism to the json there as well.

dejaloomer:
for example, here is the script to replicate the stock widget's drawing. It's basically a 1-1 lift from the cpp code. I think this could be helpful for widget developers to iterate their designs.
```
font,default
fill,black
fr,0,70,240,49,white
fr,0,111,240,20,lightgrey
fcol,white
fbcol,black
textc,52 Week,120,185,11
textc,H: $250.12,120,200,11
textc,L: $100.88,120,215,11
text,MegaCo,120,121,11,mc,black,lightgrey
ftri,110,11,130,11,120,23,red
arc,120,120,120,118,0,360,red,black
text,12%,120,48,29,mc,red,black
text,MGCO,120,92,29,mc,black,white
text,$288.60,120,155,29,mc,white,black
```

brett-dot-tech:
Nice I see, Im going to leave this PR open until I have some time to dig deeper into it, and ultimately I think we'll need to figure out how it exists along the web data widget whether as a replacement, improvement, or if it exists as a separate thing.

With how limited flash is we just need to take caution with redundant features; though I do really like how you've done this, I just think merging it will require some consideration/changes to webdata and I just need the post-holidays time to sit down and think that through (:

dejaloomer:
cool thanks!

sjoerdverweij:
I love this, but it really feels like a solution in search of a problem :-(

firmware/src/core/utils/DrawScript.cpp
```diff
@@ -0,0 +1,302 @@
...
+        } else if (pc == 8) {
+            m_manager.drawArc(parseInt(p[1]), parseInt(p[2]), parseInt(p[3]), parseInt(p[4]), parseInt(p[5]), parseInt(p[6]), parseColor(p[7]), parseColor(p[8]),true);
+        }
+        break;
+    case COMMAND_SMOOTH_ARC:
+        if (pc == 9) {
+            m_manager.drawSmoothArc(parseInt(p[1]), parseInt(p[2]), parseInt(p[3]), parseInt(p[4]), parseInt(p[5]), parseInt(p[6]), parseColor(p[7]), parseColor(p[8]), parseInt(p[9]));
+        }
+        break;
+    case COMMAND_IMAGE:
```
warmfusion, line 108: There is no such method as `drawFsJpg` in the main ScreenManager implementation.

```
firmware/src/core/utils/DrawScript.cpp: In member function 'void DrawScript::processDrawCommand(DrawScript::CommandType, const std::vector<String>&)':
firmware/src/core/utils/DrawScript.cpp:110:23: error: 'class ScreenManager' has no member named 'drawFsJpg'; did you mean 'drawString'?
             m_manager.drawFsJpg(parseInt(p[1]), parseInt(p[2]), p[3].c_str());
                       ^~~~~~~~~
                       drawString
compilation terminated due to -Wfatal-errors.
```

Reading @dejaloomer 's fork a bit more, i think this method was added during other local work to add `LittleFS` support - https://github.com/dejaloomer/info-orbs/commit/4ea15aa5deda857ee7388ec51225cae3203558c3

dejaloomer: the jpg changes are present on the dev branch which is what this PR is targeting. (and is my understanding of where all changes go initially?)

warmfusion: I retract my previous comment. Sorry about that @dejaloomer .

I tend to work in gitlab against main branches so i wasn't expecting the use of dev like that.

I had problems building my code which I attributed to this branch having issues, but after more learning i found my config.h based on `main` needed updating.

I'm very sorry.

dejaloomer: No worries. I'm still learning my way around GitHub myself- as shown by my accidentally syncing main into this pr branch 😀  (now repaired...I think...)

PR #201, opened 2024-12-28
New HTTP Task Management including changes to Stock and Weather widgets.

dreed47:
A new HTTPClientWrapper class.  Any widget needing to make an http request calls HTTPClientWrapper::addRequest() passing the url and a callback method.  HTTPClientWrapper is responsible for queueing the request.  The main loop calls HTTPClientWrapper::getInstance()->processRequestQueue().  HTTPClientWrapper will then check the request queue to see if there are requests to process. If there is it will attempt to get the semaphore to ensure only one requests runs at a time.  It will then create a Task for the http request.  Once the request returns the semaphore is released and the response is placed on a second response queue.  The main loop also calls HTTPClientWrapper::getInstance()->processResponseQueue().  HTTPClientWrapper checks the response queue and if there are response to process it pulls the response data and calls the callback function.    This change includes changes to the stock widget and the weather widget.  This simplifies things from a widget standpoint.  No need to worry about tasks, semaphores, queues.  Just call addRequest with your url and callback method. Then in your callback method process your http response and codes as normal. An optional preProcessResponse callback is also provided where the widgets can manipulate the http response before it's passed thru the processResponse callback.  This is useful for doing things like json filtering.  See Weather Widget for an example.


firmware/src/core/utils/HTTPClientWrapper.h
```diff
@@ -0,0 +1,57 @@
...
+        ResponseCallback callback;
+    };
+
+    struct ResponseData {
+        int httpCode;
+        String response;
+        ResponseCallback callback;
+    };
+
+    static HTTPClientWrapper* instance;
```
learn-more, line 43: If the `HTTPClientWrapper` is already more or less a singleton, why have the individual fields below also be static?
Considering there is only one instance of this class, they can be members just fine?

firmware/src/core/utils/HTTPClientWrapper.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    // Only try to take semaphore if we have work to do
+    if (xSemaphoreTake(requestSemaphore, 0) != pdTRUE) {
+        //Serial.println("⚠️ Semaphore blocked - request already in progress");
+        return;
+    }
+
+    Serial.println("✅ Obtained semaphore");
+    activeRequests++;
```
learn-more, line 73: What is the point of this?
The semaphore will guard it, so this is always one (when busy) or 0 (when not busy) ?

firmware/src/main.cpp
```diff
@@ -89,6 +89,8 @@ void loop() {
 
         MainHelper::checkCycleWidgets();
         wifiManager->process();
+        HTTPClientWrapper::getInstance()->processRequestQueue();  
+        HTTPClientWrapper::getInstance()->processResponseQueue();
```
learn-more, line 93: I'd combine those functions, that way you can guarantee the order of the handling and other things that might be important for various reasons.

firmware/src/core/utils/HTTPClientWrapper.h
```diff
@@ -0,0 +1,57 @@
...
+    static SemaphoreHandle_t requestSemaphore;
+    static QueueHandle_t requestQueue;
+    static QueueHandle_t responseQueue;
```
learn-more, line 46: There is no point in the responseQueue?
There should always be 1 or 0 responses, so either use a queue size of 1, or remove it entirely?

firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -46,42 +46,28 @@ void StockWidget::draw(bool force) {
...
-            if (xTaskCreate(taskGetStockData, "StockDataTask", 8192, this, 1, &m_taskHandle) == pdPASS) {
-                Serial.println("StockDataTask created");
-                m_stockDelayPrev = millis();
-            } else {
-                Serial.println("Failed to create StockDataTask");
-                setBusy(false);
-            }
-        } else {
-            // Serial.println("StockDataTask is already running");
+        setBusy(true);
```
learn-more, line 49: each individual orb calling `setBusy` does not make sense, especially when it is no longer blocking.
Instead, the request wrapper should do this, calling `setBusy(true)` when it starts handling a request, and `setBusy(false)` when the request is handled (or something along those lines)

---

The whole thing where there are multiple code paths trying to control one resouce (a led) already did not make sense before, but now it is even worse imo.

firmware/src/core/utils/HTTPClientWrapper.h
```diff
@@ -0,0 +1,57 @@
+#ifndef HTTP_CLIENT_WRAPPER_H
+#define HTTP_CLIENT_WRAPPER_H
+
+#include <Arduino.h>
+#include <HTTPClient.h>
```
learn-more, line 5: This `HTTPClient.h` does not have to be included here, it is an implementation detail of the `HTTPClientWrapper`

firmware/src/core/utils/HTTPClientWrapper.cpp
```diff
@@ -0,0 +1,167 @@
...
+SemaphoreHandle_t HTTPClientWrapper::requestSemaphore = nullptr;
+QueueHandle_t HTTPClientWrapper::requestQueue = nullptr;
+QueueHandle_t HTTPClientWrapper::responseQueue = nullptr;
+
+// Initialize static debug counters
+volatile uint32_t HTTPClientWrapper::activeRequests = 0;
+volatile uint32_t HTTPClientWrapper::maxConcurrentRequests = 0;
+
+HTTPClientWrapper::HTTPClientWrapper() {
+    if (!requestSemaphore) {
```
learn-more, line 13: Additionally, if these are just members of the `HTTPClientWrapper` there is no need to do these checks both here and in the destructor. (That need already wasn't there btw, since it is a singleton)

learn-more (commented):
Some points where I think the PR could be improved.

firmware/src/core/utils/HTTPClientWrapper.cpp
dreed47: correct.  it was just a double check while debugging to ensure that only one at max

firmware/src/main.cpp
dreed47: Combining them won't guarantee the order of handling since the processRequestQueue() generates a non-blocking Task thread.  You can't wait for the response or it would be blocking.

firmware/src/core/utils/HTTPClientWrapper.h
dreed47: I disagree.  The request is spun off on a Task thread that is independent of the main loop.  Yes only one can run at a time but since we're processing the callback from the main loop in theory it could be possible to get one response stepping on another.  If we DON'T process the callback in the main loop then we could eliminate the responseQueue but that also means your processing response data and updating things OUTSIDE the main loop that could cause race conditions.   The responseQueue limits that risk as it allows us to keep widget processing in the main loop.

firmware/src/widgets/stockwidget/StockWidget.cpp
dreed47: yeah I agree

firmware/src/core/utils/HTTPClientWrapper.h
dreed47: not sure I understand...this is the implementation of HTTPClientWrapper.

firmware/src/core/utils/HTTPClientWrapper.cpp
dreed47: yeah. destructor never gets used since it a singleton and lives with the app

firmware/src/core/utils/HTTPClientWrapper.h
dreed47: They could be member variables instead, I've always used static variables with a singleton as I never have to think about whether the data is shared or not.  In this case there's ever only one so it doesn't matter as much

firmware/src/core/utils/HTTPClientWrapper.h
learn-more: and who of the _users_ of the `HTTPClientWrapper` needs to know about the `HTTPClient` ?

firmware/src/widgets/stockwidget/StockWidget.cpp
dreed47: I created a Utils::setBusy()

firmware/src/core/utils/HTTPClientWrapper.cpp
dreed47: destructor removed

PR #206, opened 2024-12-31
Move taskSemaphore to GlobalResources

dreed47:
The purpose of the taskSemaphore is to ensure at most only one non-mainline Task thread can run at a time to conserver/manage resource.  Moving it a GlobalResources allows it to be used easier anywhere in the app.  I also included comments as to the semaphore usage.


dreed47:
will reroll with the renaming of HTTPClientWrapper PR

PR #207, opened 2025-01-02
Enable watchdog timer

flattermann:


learn-more:
Did you observe this working in practice?

My experience in general with a watchdog is that only an 'external' watchdog is reliable.
e.g. esp watching itself -> if the thing crashes, this won't happen. (And this _should_ be the general use case?)

flattermann:
I will do some more tests, but yes, it did trigger here.

Sometimes the HTTP request and especially the ArduinoJson parsing is freezing my ESP and this successfully triggered a restart, but I don't know if it works 100%. At least, it shouldn't have negative side effects, I think.

PR #208, opened 2025-01-02
Don't block when trying to connect to mqtt server

dejaloomer:
Current code stays in a loop until it connects to mqtt server which can hang the device. This retries on interval.

### DIFF
--- a/firmware/src/widgets/mqttwidget/MQTTWidget.cpp
+++ b/firmware/src/widgets/mqttwidget/MQTTWidget.cpp
@@ -340,13 +340,21 @@ void MQTTWidget::subscribeToOrbs() {
     }
 }
 
+#define RECONNECT_INTERVAL 5000
 // Handle MQTT reconnection
 void MQTTWidget::reconnect() {
     //    Serial.println("Inside reconnect method");
 
     // Loop until reconnected
-    while (!mqttClient.connected()) {
+    if (!mqttClient.connected()) {
+        // Check if it's time to attempt a reconnection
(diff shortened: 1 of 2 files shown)


dreed47 (changes requested):
I submitted comments on PR 208.  Ill copy them here for reference. "While removing this delay(500) helps it doesn't solve the blocking issue. PubSubClient is inherently a blocking library. To make it non-blocking you either need to make it true async or implement Tasks. And in either case you should refrain from making more than one non-main line threads at time as the limited resources of the device would not support it. That means you need to be aware of other non-mainline threads that might be active which means you need to inspect the global taskSemaphore. Ideally the PupSubClient should be wrapped similar to how the HTTPClient is being wrapped and ideally in the same task management code.  We also need to be careful to not be making updates to widgets outside of the mainline loop"

dejaloomer:
Other areas of pubsub or network stack may block for a while but they will eventually time out and return. The current code is guaranteed to hang perpetually while there is any sort of issue connecting to server- network layer, auth issue, server down, etc. This PR eliminates that class of problem.

dreed47:
> o

yes but this is a band-aid at best.  It's fine if we merge this fix but for the app to be stable and response we have to get all blocking functions removed and I would not want anyone to assume that this fix does that as it does not.

brett-dot-tech:
LGTM, going to merge will note in an issue that this issue still exists, I dont see any harm in merging this as a bandaid like you say though in the meantime.

PR #209, opened 2025-01-03
Publish a configurable topic when buttons are pressed

dejaloomer:
optionally use "buttontopic" from the config message to provide a button-press topic. On button press the topic is sent a dictionary with two keys:
"button" = "invalid", "left", "middle", or "right"
"state" = "nothing", "short", "medium", or "long"


brett-dot-tech:
This is great;

Just to make sure as a non MQTT user, this effectively would let you use your orbs as a MQTT input device, yes? to control lights and other such devices?

dejaloomer:
Correct. I use Home Assistant to listen for the button topic and trigger automations with it.

PR #210, opened 2025-01-03
Show Stock change as Price or Percent

HugoBalder:
Option added to switch between percent or price change for stocks, including the currency of the stock. This option is also available in the advanced section of the Stock Widget configuration in the WifiManager.

### DIFF
--- a/firmware/src/widgets/stockwidget/StockWidget.cpp
+++ b/firmware/src/widgets/stockwidget/StockWidget.cpp
@@ -12,7 +12,7 @@ StockWidget::StockWidget(ScreenManager &manager, ConfigManager &config) : Widget
 
     String stockchangeFormats[] = {"Percent", "Price"};
     m_config.addConfigComboBox("StockWidget", "stockchgFmt", &m_stockchangeformat, stockchangeFormats, 2, "Show percent or price change", true);
-        
+
     char *symbol = strtok(stockList, ",");
     m_stockCount = 0;
     do {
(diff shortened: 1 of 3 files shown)


flattermann:
Can you please make sure that your IDE uses the correct code formatting? (There is a file `.clang-format` that defines the code format.)

For VSCode, just open the info-orbs.code-workspace file or enable format-on-save in preferences, and it should automatically format your files.

firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -131,7 +134,11 @@ void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint3
         m_manager.fillTriangle(110 + arrowOffsetX, 132 + arrowOffsetY, 130 + arrowOffsetX, 132 + arrowOffsetY, 120 + arrowOffsetX, 120 + arrowOffsetY, TFT_GREEN);
         m_manager.drawArc(centre, centre, 120, 118, 0, 360, TFT_GREEN, TFT_GREEN);
     }
-    m_manager.drawString(stock.getPercentChange(2) + "%", centre, 48, bigFontSize, Align::MiddleCenter);
+    if (!m_stockchangeformat) {
+        m_manager.drawString(stock.getPercentChange(2) + "%", centre, 48, bigFontSize, Align::MiddleCenter);
+    } else {
+        m_manager.drawString(stock.getPriceChange(2) + stock.getCurrencySymbol(), centre, 48, bigFontSize, Align::MiddleCenter);
```
lyricnz, line 140: The currency symbol for the High/Low price is **before** the value, and here it's **after**.  Let's be consistent (even if some locale would do it differently, let's not open that can of worms).

Likewise. the hardcoded "2" (decimals) is inconsistent with the high/low price (which looks like it uses the data straight from the API). Should we just do the same?

Disclaimer: code not tested.

lyricnz (approved):

PR #212, opened 2025-01-04
Pin platform and Arduino versions

HeMan:
Code fails to compile on newer versions of Arduino and platform


learn-more (changes requested):
I would suggest to fix the issues that occur instead of hiding them,
this will bite us later.

HeMan:
> I would suggest to fix the issues that occur instead of hiding them,
> this will bite us later.

Depending on if the dependencies support Arduino 3.x it could become a somewhat bigger project to fix it. There have also been some controversy wrt Platformio an Arduino 3, https://www.cnx-software.com/2024/06/01/espressif-releases-arduino-esp32-core-3-0-0-but-platformio-support-in-doubt/ ,that could be worth taking in consideration.

flattermann:
Just a quick question, not meant to sound negative, but who actually needs this change?

Is it necessary if you have newer PlatformIO/Arduino libs installed?

Because apparently, if you install PlatformIO from scratch (like e.g. the github action does), it seems to work fine without restricting the version.

Also, please use platform version 6.9.0+ (that's the version that will be used if installed from scratch):

`PLATFORM: Espressif 32 (6.9.0) > DOIT ESP32 DEVKIT V1`

lyricnz:
@HeMan can you describe how to reproduce the problem with "fails to compile"?

I just opened a fresh checkout of <dev> on a brand new machine, without PlatformIO installed.  I installed PlatformIO, then let it do "all the things", which took a couple of minutes.  I then copied `firmware/config/config.h.template` to `config.h` and clicked the "build" button in the project task, which looks like it worked.

image

HeMan:
I did some more testing, and the problem only arises when you have done other Platform.io projects that uses Arduino 3.x, so it's a really small group that gets bitten by this. However, having pinned versions is nice to have, since it ensures that every build uses the same Arduino versions.
I'll update the PR to use the latest platform version.

PR #215, opened 2025-01-05
Fix: Small bug in WebDataElementCircleModel Parser

oecheverri:
Fixes a small bug in the WebDataElementCircleParser for the 'filled' field as it was checking for an integer value, where instead it should be looking for a boolean.


ericthelin (dismissed):

lyricnz (approved):

flattermann (approved):

flattermann (changes requested):
Please change the target branch to dev.

oecheverri:
@flattermann done!

PR #216, opened 2025-01-06
Memory debug interval

dreed47:
Provides a way to define the frequency of memory debug updates

### DIFF
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -130,7 +130,6 @@
 
 #define MAX_RETRIES 3
 
-// #define MEMORY_DEBUG true
 // #define MEMORY_DEBUG_INTERVAL 5000
 
 #endif
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -130,4 +130,7 @@
 
 #define MAX_RETRIES 3
 
+// #define MEMORY_DEBUG true
+// #define MEMORY_DEBUG_INTERVAL 60000
+
 #endif
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -131,6 +131,6 @@
 #define MAX_RETRIES 3
 
 // #define MEMORY_DEBUG true
-// #define MEMORY_DEBUG_INTERVAL 60000
+// #define MEMORY_DEBUG_INTERVAL 5000
 
 #endif
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -132,6 +132,7 @@
 
 #define MAX_RETRIES 3
 
+// if not defined, then no memory debugging is emitted
 // #define MEMORY_DEBUG_INTERVAL 5000
 
 #endif
--- a/firmware/src/core/utils/ShowMemoryUsage.cpp
+++ b/firmware/src/core/utils/ShowMemoryUsage.cpp
@@ -6,8 +6,9 @@ static unsigned long s_lastMemoryUsageShownAt = 0;
 
 void ShowMemoryUsage::printSerial(bool force, bool newLine) {
     multi_heap_info_t info;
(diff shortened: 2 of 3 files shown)


firmware/config/config.h.template
```diff
@@ -130,4 +130,7 @@
 
 #define MAX_RETRIES 3
 
+// #define MEMORY_DEBUG true
```
lyricnz, line 133: MEMORY_DEBUG is unused?

firmware/config/config.h.template
```diff
@@ -130,4 +130,7 @@
...
+// #define MEMORY_DEBUG_INTERVAL 5000
```
lyricnz, line 134: Could/should we combine these?  If MEMORY_DEBUG_INTERVAL is defined then enable+use it, otherwise disable memory_debug?

lyricnz: Thanks. Maybe a comment saying that if this value is not defined, then no memory debugging is emitted?

firmware/src/core/utils/ShowMemoryUsage.cpp
```diff
@@ -1,15 +1,22 @@
...
+#include "../../../config/config.h"
 #include <Arduino.h>
 
 // initialize static members
 static unsigned long s_lastMemoryUsageShownAt = 0;
 
+#ifdef MEMORY_DEBUG_INTERVAL
+    static const unsigned long interval = MEMORY_DEBUG_INTERVAL;
+#else
+    static const unsigned long interval = 1000;
```
lyricnz, line 11: This value is actually unused (when ndef) - maybe we can use a less confusing value?

sjoerdverweij: Add a unit? interval_ms?

lyricnz: @dreed47 I actually meant that using a default (when ifndef MEMORY_DEBUG_INTERVAL) value of `1000` is confusing - because the value is not used (and doesn't matter), so it's not "1000" anything.

lyricnz: Maybe `0` or `-1` - with a comment saying "unused"?

lyricnz:
Thought: should the memory debugging include a timestamp?  Otherwise it's hard to know (retrospectively) how big an issue the memory usage was...

firmware/src/core/utils/ShowMemoryUsage.cpp
dreed47: Ahh, yeah. You're correct.  It mattered when the PR was using 2 config entries but not when there's only one.  I'll set it to -1 with a comment

firmware/src/core/utils/ShowMemoryUsage.cpp
```diff
@@ -1,15 +1,22 @@
 #include "ShowMemoryUsage.h"
+#include "../../../config/config.h"
```
learn-more, line 2: This should not be needed.

firmware/src/core/utils/ShowMemoryUsage.cpp
```diff
@@ -1,15 +1,22 @@
...
 #include <Arduino.h>
 
 // initialize static members
 static unsigned long s_lastMemoryUsageShownAt = 0;
 
+#ifdef MEMORY_DEBUG_INTERVAL
+static const unsigned long interval_ms = MEMORY_DEBUG_INTERVAL;
+#else
+static const unsigned long interval_ms = 1000; // default is needed for cases of where this method is called without MEMORY_DEBUG_INTERVAL being set
```
learn-more, line 11: I'd use the same 5000 here

learn-more (approved):

firmware/src/core/utils/ShowMemoryUsage.cpp
dreed47: changed

lyricnz (approved):

PR #218, opened 2025-01-06
Rename HTTPClientWrapper to Taskmanager

dreed47:
- Renamed HTTPClientWrapper to the more generic TaskManager
- Added TaskFactory for creating task instances
- Moved taskSemaphore to util/GlobalResources
- Modified Weather and Stock widget to use TaskManager to perform it's http requests

PR #219, opened 2025-01-06
Super-rough structure-only per-screen refactor

sjoerdverweij:


firmware/src/core/Feed.h
```diff
@@ -0,0 +1,22 @@
...
+    FeedStock = 2,
+    FeedMQTT = 3,
+    MAXFEEDID = 4
+};
+
+struct Feed {
+public:
+   virtual String& getName();
+   virtual FeedId getId();
+   template<class DerivedFeedType> DerivedFeedType getFeed() { return *dynamic_cast<DerivedFeedType*>(this); }
```
learn-more, line 19: I would suggest not to use a `dynamic_cast`.
A `static_cast` + a check on the correct type (id) would be a lot faster.

A `dynamic_cast` is a very slow function, that _requires_ a certain amount of debug info to be present in the binary.

firmware/src/core/Feeds.h
```diff
@@ -0,0 +1,25 @@
+#ifndef FEEDS_H
+#define FEEDS_H
+
+#include <Arduino.h>
+#include "Feed.h"
+#include "..\..\feeds\FeedFactory.h"
```
learn-more, line 6: You need to use forward slashes, linux chokes on the backslash, windows does not care so both work there.

firmware/src/core/Feed.h
```diff
@@ -0,0 +1,22 @@
...
+    FeedDateTime = 0,
+    FeedWeather = 1,
+    FeedStock = 2,
+    FeedMQTT = 3,
+    MAXFEEDID = 4
+};
+
+struct Feed {
+public:
+   virtual String& getName();
```
learn-more, line 17: Is the name required?
I'd say that 'id' would be enough.

and a nit: this is indented with 3 spaces.

firmware/src/core/Feed.h
```diff
@@ -0,0 +1,22 @@
...
+   virtual FeedId getId();
```
learn-more, line 18: No need to make it virtual, just have a `const FeedId Id` in the class, and create a constructor that you need to pass the `FeedId` into:

```cpp
struct Feed {
    const FeedId Id;
    explicit Feed(FeedId id)
        : Id(id)
    {
    }
};
```

firmware/src/core/Feed.h
```diff
@@ -0,0 +1,22 @@
...
+{
+    FeedDateTime = 0,
+    FeedWeather = 1,
+    FeedStock = 2,
+    FeedMQTT = 3,
+    MAXFEEDID = 4
+};
+
+struct Feed {
+public:
```
learn-more, line 16: `public` is not needed in a struct as first item, everything in a struct is public by default

firmware/src/core/Feeds.h
```diff
@@ -0,0 +1,25 @@
...
+struct Feeds {
+private:
+  Feed* _Feeds[MAXFEEDID];
+public:
+  Feed& registerFeed(const FeedId feedId) {
+    if (!_Feeds[feedId])
+    {
+        _Feeds[feedId] = FeedFactory::CreateFeed(feedId);
+    }
+  }
```
learn-more, line 17: You should return something here, or remove the return type

firmware/src/core/Feeds.h
```diff
@@ -0,0 +1,25 @@
...
+
+  Feed& getFeed(const FeedId feedId) {
+    assert(_Feeds[feedId]);
```
learn-more, line 20: This suggests that all feeds are always created, I would advice against that.
Anything that is not needed, does not have to be loaded.

firmware/src/core/Orb.h
```diff
@@ -0,0 +1,25 @@
...
+#include "Renderer.h"
+
+enum OrbId
+{
+    OrbDateTimeWeekday = 1,
+    OrbCityWeatherPrediction = 2,
+    OrbWeatherIcon = 3,
+    OrbTemperatureHighLow = 4,
+    OrbWeather3DayForecast = 5
+};
```
learn-more, line 16: Something like this allows to mix and match the orbs, which is a long-standing feature request:

```cpp
enum class OrbId
{
    Invalid,

    Weather_DateTimeWeekday,
    Weather_CityPrediction,
    Weather_Icon,
    Weather_TemperatureHighLow,
    Weather_3DayForecast,

    Stocks_1,
    Stocks_2,
    Stocks_3,
    Stocks_4,
    Stocks_5,

    Clock_Hour_h,
    Clock_Hour_l,
    Clock_SecondsColon,
    Clock_Minute_h,
    Clock_Minute_l,
};
```

dreed47:
I like the overall design direction.  I have a couple logical model recommendations.

1. A Gadget represents 5 Orbs but as I see it it needs to support both pre-defined Orb sets like clock and also mix/match user defined Orb sets.  I'd recommend abstracting out a ContentSet class.  Then maybe rename Gadget to ContentManager.  So the ContentManager manages a collection of ContentSets and allow switching between them.  So the ContentManager manages both predefined and user-defined ContentSets, handle button interations to cycle through the ContentSets and provides methods for creating, updating, and saving user-defined ContentSets. 

2. Feeds - The ContentManager (Gadget) should manage the lifecycle of the Feeds, ensure that Feeds are shared across all Orbs to reduce redundant data fetching, and dynamically load Feeds when needed by one or more Orbs.  Also Feeds should make use of TaskManager/TaskFactory when getting data from external APIs

3. Button Handling - ContentManager (Gadget) should specifically manage cycling of ContentSets via left/right button clicks.

4. Renderer - The ContentManager (gadget) should coordinate with the Renderer to draw the active ContentSet on the 5 displays.  Each Orb in the active ContentSet should use the Renderer to draw its content on the appropriate display.

5. Need support for persistence for User-Defined ContentSets?  to non-volatile storage?  then would need methods for loading saved ContentSets on startup

sjoerdverweij:
> I like the overall design direction. I have a couple logical model recommendations.
> 
> 1. A Gadget represents 5 Orbs but as I see it it needs to support both pre-defined Orb sets like clock and also mix/match user defined Orb sets.  I'd recommend abstracting out a ContentSet class.  Then maybe rename Gadget to ContentManager.  So the ContentManager manages a collection of ContentSets and allow switching between them.  So the ContentManager manages both predefined and user-defined ContentSets, handle button interations to cycle through the ContentSets and provides methods for creating, updating, and saving user-defined ContentSets.
> 2. Feeds - The ContentManager (Gadget) should manage the lifecycle of the Feeds, ensure that Feeds are shared across all Orbs to reduce redundant data fetching, and dynamically load Feeds when needed by one or more Orbs.  Also Feeds should make use of TaskManager/TaskFactory when getting data from external APIs
> 3. Button Handling - ContentManager (Gadget) should specifically manage cycling of ContentSets via left/right button clicks.
> 4. Renderer - The ContentManager (gadget) should coordinate with the Renderer to draw the active ContentSet on the 5 displays.  Each Orb in the active ContentSet should use the Renderer to draw its content on the appropriate display.
> 5. Need support for persistence for User-Defined ContentSets?  to non-volatile storage?  then would need methods for loading saved ContentSets on startup

Gadget only needs to support configurable mix/match orb sets, because there really no longer is such a thing as a predefined widget; those are just default mix/match configurations. You'd start out with a configuration for a Gadget that has 4 instances of a ClockDigit orb, and one ClockSeconds orb. If we want to prevent users from changing the default settings, we can make them read-only -- but they will still work the exact same way.
Lifecycle and deduplication of Feeds is in the Feeds class. It should be a singleton, which it currently is not properly set up and referenced as.
Button handling etc.; I think Gadgets is what you want ContentManager to be.
Agreed, it will, the draw code isn't fleshed out yet.
I'm open to the best way to save settings. My current thinking is Serialize/Deserialize methods on the top of the settings hierarchy (Settings) which (de)serializes the global settings, then calls all its GadgetSettings, which then calls all its OrbSettings. (GadgetSettings holds which orbs a gadget shows, OrbSettings has the specific settings for that instance of the orb).

firmware/src/core/Feed.h
sjoerdverweij: I forget that all the time, so I like to be explicit. If we decide it's an enforced style thing, no problem, I'll take it out.

firmware/src/core/Feed.h
sjoerdverweij: We'll need some kind of name to show in the config UI. Whether this is the best way to do it is debatable, I am open to suggestions.

firmware/src/core/Orb.h
sjoerdverweij: Yes, that's the point of it, I just started with the weather widget -- are you saying you don't like the naming of the enum members?

firmware/src/core/Feed.h
sjoerdverweij: Good point, done.

firmware/src/core/Feed.h
sjoerdverweij: Good point, done. Not even going to check on type, this would be very much caught during dev.

firmware/src/core/Orb.h
learn-more: I guess so.

In cases like this, having a consistent style in naming pays off imo.

lyricnz:
Sorry I broke this PR merging #218

firmware/src/core/Orb.h
lyricnz: Why would we need an enum?  Why not a list of list-of instances?  (with constuctor/factory parameters as required). Totally value pseudo-code

```cpp
[
  [
    new Clock(mode=HourH),
    new Clock(mode=HourL),
    new Clock(mode=SecondsColon),
    new Clock(mode=SecondsH),
    new Clock(mode=SecondsL)
  ],
  [
    new Stock(symbol=whatever, currency=whatever, apikey=whateever),
    ...
  ],
  [
    new WeatherTemp(..., location=whatever),
    new WeatherRain(..., location=whatever),
    new WeatherThing(..., location=whatever),
    ...
  ]
]
```

lyricnz: I guess we need something in the UI, ya

learn-more: In the UI, and in the config files / web editor / whatever is used to configure this thing.
You can't really transmit class instances over the web ;)

Issue #203, opened 2024-12-30
Implement ArduinioLog for better log message control


dreed47:
t's pretty lightweight and it would give us the ability to better control what gets printed to the Serial Monitor.  It supports these log types.

```
// Log levels: 0 (SILENT), 1 (FATAL), 2 (ERROR), 3 (WARNING),
// 4 (NOTICE), 5 (TRACE), 6 (VERBOSE)
#define LOG_LEVEL LOG_LEVEL_VERBOSE
```

So instead of 
`Serial.println("Starting up...");`
  It would be something like 
`Log.noticeln("Starting up...");`

For reference here's the [github page for ArduinoLog](https://github.com/thijse/Arduino-Log).

PR #220, opened 2025-01-07
Implement ArduinioLog for better log message control 

dreed47:
Per feature request in issue #203 this change enables the use to AduinoLog to allow better control over log output.  The change only converts Serial.print to Log format in main.cpp and MainHelper.cpp as making global changes would affect almost every cpp file in the system.  Merging this change first will allow other modules to be changed gradually over time as files are updated.


sjoerdverweij (approved):

learn-more (approved):

lyricnz:
@dreed47 might be worth raising PRs to change other output.  I'd be interested in adding timestamps to the output also.

PR #228, opened 2025-01-16
Include missing config_helper.h in ShowMemoryUsage

flattermann:
Added the missing config_helper.h include to resolve potential dependency issues. This ensures that all necessary configurations are properly loaded for memory usage functionality.


learn-more:
The line `-include "firmware/config/config_helper.h"` in `platformio.ini` forces this include to be added to everything automatically, so it does not have to be included manually.

flattermann:
Well, that didn't work for me.

If I add
`#define MEMORY_DEBUG_INTERVAL 1000`
in my config.h, it would print the memory debug every 5000ms.

When I add the `#include` statement, it works.

PR #229, opened 2025-01-16
Refine JSON filter and update portfolio value handling.

flattermann:
Updated the JSON deserialization filter to include more precise fields for holdings and performance metrics. Adjusted how portfolio values are processed to align with the refined filter structure for accuracy and consistency.


flattermann:
Superseded by #236.

PR #231, opened 2025-01-28
Added new Clocktype Flipclock with animation

HugoBalder:
Real Flipclock animation supporting 12h/24h modes


lyricnz:
That's a whole lotta JPG. Do we have space?

HugoBalder:
After uploading all that stuff to my ESP i have with 93% used flash.

lyricnz:
Maybe it should be disabled by default then?

HugoBalder:
Disabled it by default in the config.h.template file. This saves 201kb of flash memory.

flattermann:
I really like the idea of having animated clocks. But I'd prefer if we would have some general way to animate custom clocks.

This would allow us to also share the animated clocks in the future "clock repo website" that @brett-dot-tech is planning.

For this to work, we would need to
- load the images from LittleFS instead of Flash (we have to check if that is fast enough)
- use a common naming schema for the files (I would recommend something like X-Y-Z.jpg with X=start digit, Y=end digit, Z=step)
- rename colon off/on to 10/11 like in the custom clocks
- find a way to recognize if a custom clock is animated (the easiest way would be a setting in the WebUI for now, but maybe we could also just check for a special (empty?) file ("isAnimated") in the custom clock directory
- check if another step exists in the current animation (e.g. just check if the respective X-Y-Z.jpg exists), if yes: show it, if not, keep the current frame until the digit changes

This is probably quite a bit of work, but then we could use the same code for multiple clocks...

HugoBalder:
As nobody is interested in this flipclock anymore, i close and delete this PR.

PR #236, opened 2025-02-19
ParqetWidget: Use a proxy to reduce JSON size

flattermann:
This will drastically reduce the memory footprint on the ESP (like 100kB -> 3kb) and increase the parsing speed.

This also allows us to use `TaskFactory::createHttpGetTask` to fetch the Parqet data in background.

I'm hosting the proxy on my server at the moment, but everybody can host their own proxy if they want (e.g. for privacy reasons).

The proxy source code (in Python) with Docker support can be found here: 
https://github.com/flattermann/info-orbs-parqet-proxy/

This PR supersedes #229.

### DIFF
--- a/firmware/src/widgets/parqetwidget/ParqetHoldingDataModel.cpp
+++ b/firmware/src/widgets/parqetwidget/ParqetHoldingDataModel.cpp
@@ -5,68 +5,86 @@
 ParqetHoldingDataModel::ParqetHoldingDataModel() {
 }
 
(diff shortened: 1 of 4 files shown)


flattermann:
I just tried this Parqet portfolio with 71 assets (18 widget pages) and it worked fine:

https://app.parqet.com/p/6756061197b522e6d882e465

The proxied JSON is around 15kB instead of >400kB. Nice. 🙂

PR #239, opened 2025-02-21
Orbs screen saver - Matrix rain style

nconner-06:
A small widget to provide a 'Matrix' style rain effect..

I think I've included everything..

Enjoy.

### DIFF
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -15,6 +15,7 @@
 #define INCLUDE_PARQET        // include ParqetWidget
 //#define INCLUDE_WEBDATA       // include WebDataWidget
 //#define INCLUDE_MQTT          // include MQTTWidget
+#define INCLUDE_SCREENSAVER
 
 // NIGHTTIME DIMMING
 //#define DIM_START_HOUR 22  // Dim the screens at this time (24h format)
--- a/firmware/src/core/screenmanager/ScreenManager.cpp
+++ b/firmware/src/core/screenmanager/ScreenManager.cpp
@@ -18,6 +18,13 @@ ScreenManager::ScreenManager(TFT_eSPI &tft) : m_tft(tft) {
     m_tft.fillScreen(TFT_WHITE);
     m_tft.setTextDatum(MC_DATUM);
     reset();
+    #ifdef INCLUDE_SCREENSAVER
+        ConfigManager *cm = ConfigManager::getInstance();
+        bool l_smallFont = cm->getConfigBool("SmallFont", false);
+        if (l_smallFont) m_matrix_effect.init(&tft, true, false);
+        else m_matrix_effect.init(&m_tft);
+    #endif
+
 
     // Init TJpg_Decode
     TJpgDec.setSwapBytes(true); // JPEG rendering setup
--- a/firmware/src/core/screenmanager/ScreenManager.h
+++ b/firmware/src/core/screenmanager/ScreenManager.h
@@ -8,6 +8,7 @@
 #include <SPI.h>
 #include <TFT_eSPI.h>
 #include <TJpg_Decoder.h>
+#include "ScreenSaverwidget/DigitalRainAnimation.hpp"
 
 #define NUM_SCREENS 5
 
@@ -83,6 +84,8 @@ public:
     static bool tftOutput(int16_t x, int16_t y, uint16_t w, uint16_t h, uint16_t *bitmap);
     JRESULT drawJpg(int32_t x, int32_t y, const uint8_t jpeg_data[], uint32_t data_size, uint8_t scale = 1, uint32_t imageColor = 0);
     JRESULT drawFsJpg(int32_t x, int32_t y, const char *filename, uint8_t scale = 1, uint32_t imageColor = 0);
+
+    DigitalRainAnimation<TFT_eSPI> m_matrix_effect = DigitalRainAnimation<TFT_eSPI>();
 
 private:
     static ScreenManager *instance;
--- a/firmware/src/main.cpp
+++ b/firmware/src/main.cpp
@@ -7,9 +7,11 @@
 #include "weatherwidget/WeatherWidget.h"
 #include "webdatawidget/WebDataWidget.h"
 #include "wifiwidget/WifiWidget.h"
+#include "ScreenSaverwidget/ScreenSaverwidget.h"
 #include <ArduinoLog.h>
 
 TFT_eSPI tft = TFT_eSPI();
+DigitalRainAnimation<TFT_eSPI> matrix_effect = DigitalRainAnimation<TFT_eSPI>();
 
 GlobalTime *globalTime{nullptr};
 WifiWidget *wifiWidget{nullptr};
@@ -40,6 +42,9 @@ void addWidgets() {
 #endif
 #ifdef INCLUDE_MQTT
     widgetSet->add(new MQTTWidget(*sm, *config));
+#endif
+#ifdef INCLUDE_SCREENSAVER
+    widgetSet->add(new ScreenSaverwidget(*sm, *config));
 #endif
 }
 
--- a/firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
+++ b/firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
@@ -1,0 +1,52 @@
+#include "ScreenSaverwidget.h"
+
+ScreenSaverwidget::ScreenSaverwidget(ScreenManager &manager, ConfigManager &config) : Widget(manager, config){
+    m_enabled = true; 
+    m_config.addConfigBool("ScreenSaverWidget", "SmallFont", &m_smallFont, "Small font (unchecked)", false);
+    m_config.addConfigColor("ScreenSaverWidget", "TextColor", &m_textColor, "Text Color", false);
+    m_config.addConfigColor("ScreenSaverWidget", "HeadTextColor", &m_headtextColor, "Head Text Color", false);
+    m_config.addConfigInt("ScreenSaverWidget", "lineMin", &m_lineMin, "Minimum line length", true);
+    m_config.addConfigInt("ScreenSaverWidget", "lineMax", &m_lineMax, "Maximum line length", true);
+    m_config.addConfigInt("ScreenSaverWidget", "speedMin", &m_speedMin, "Minimum line speed", true);
+    m_config.addConfigInt("ScreenSaverWidget", "speedMax", &m_speedMax, "Minimum line speed", true);
+    m_config.addConfigInt("ScreenSaverWidget", "updateInterval", &m_updateInterval, "Screen Update Interval", true);
+}
+
+void ScreenSaverwidget::addConfigToManager(){
+}
+
+ScreenSaverwidget::~ScreenSaverwidget() {
+}
+
+void ScreenSaverwidget::setup(){
+    int R;
+    int G;
+    int B;
+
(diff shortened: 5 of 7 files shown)


nconner-06:
I've corrected the small error - lower case "w" :-(

flattermann:
Thank you for your contribution!

I'll to review the code soon, but here are some general thoughts that came to mind:

I think the Widget should be called MatrixWidget or similar, not ScreenSaverWidget, because it's just a normal Widget and not really a screensaver.

Also, if we were to implement a screensaver mechanic, I think it would be best, if the user could configure a Widget of his choice that will be displayed after a timeout.

firmware/config/config.h.template
```diff
@@ -15,6 +15,7 @@
 #define INCLUDE_PARQET        // include ParqetWidget
 //#define INCLUDE_WEBDATA       // include WebDataWidget
 //#define INCLUDE_MQTT          // include MQTTWidget
+#define INCLUDE_SCREENSAVER
```
flattermann, line 18: If you rename the Widget, please also rename the config name and add a comment with the actual widget name

firmware/src/core/screenmanager/ScreenManager.cpp
```diff
@@ -18,6 +18,13 @@ ScreenManager::ScreenManager(TFT_eSPI &tft) : m_tft(tft) {
     m_tft.fillScreen(TFT_WHITE);
     m_tft.setTextDatum(MC_DATUM);
     reset();
+    #ifdef INCLUDE_SCREENSAVER
```
flattermann, line 21: I don't think that it's a good idea to use widget specific code in ScreenManager.

The manager classes (ScreenManager, ConfigManager...) should not have knowledge of the widgets.

firmware/src/core/screenmanager/ScreenManager.h
```diff
@@ -8,6 +8,7 @@
 #include <SPI.h>
 #include <TFT_eSPI.h>
 #include <TJpg_Decoder.h>
+#include "ScreenSaverwidget/DigitalRainAnimation.hpp"
```
flattermann, line 11: ScreenManager should not include widget headers

firmware/src/main.cpp
```diff
@@ -7,9 +7,11 @@
 #include "weatherwidget/WeatherWidget.h"
 #include "webdatawidget/WebDataWidget.h"
 #include "wifiwidget/WifiWidget.h"
+#include "ScreenSaverwidget/ScreenSaverwidget.h"
 #include <ArduinoLog.h>
 
 TFT_eSPI tft = TFT_eSPI();
+DigitalRainAnimation<TFT_eSPI> matrix_effect = DigitalRainAnimation<TFT_eSPI>();
```
flattermann, line 14: Why do you need to define the `matrix_effect` here?

main.cpp should only include widget code that's absolutely necessary, i.e. `widgetSet -> add(new XXX)`

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
```diff
@@ -0,0 +1,52 @@
+#include "ScreenSaverwidget.h"
+
+ScreenSaverwidget::ScreenSaverwidget(ScreenManager &manager, ConfigManager &config) : Widget(manager, config){
+    m_enabled = true;
```
flattermann, line 4: Widgets should have an enable option in the WebUI, and they should be disabled by default.

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
```diff
@@ -0,0 +1,52 @@
...
+    m_config.addConfigColor("ScreenSaverWidget", "TextColor", &m_textColor, "Text Color", false);
+    m_config.addConfigColor("ScreenSaverWidget", "HeadTextColor", &m_headtextColor, "Head Text Color", false);
+    m_config.addConfigInt("ScreenSaverWidget", "lineMin", &m_lineMin, "Minimum line length", true);
+    m_config.addConfigInt("ScreenSaverWidget", "lineMax", &m_lineMax, "Maximum line length", true);
+    m_config.addConfigInt("ScreenSaverWidget", "speedMin", &m_speedMin, "Minimum line speed", true);
+    m_config.addConfigInt("ScreenSaverWidget", "speedMax", &m_speedMax, "Minimum line speed", true);
+    m_config.addConfigInt("ScreenSaverWidget", "updateInterval", &m_updateInterval, "Screen Update Interval", true);
+}
+
+void ScreenSaverwidget::addConfigToManager(){
```
flattermann, line 15: What's this function used for?

Did you want to move the `m_config.addConfig*()` from the constructor here?

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.h
```diff
@@ -0,0 +1,31 @@
...
+    void draw(bool force) override;
+    void buttonPressed(uint8_t buttonId, ButtonState state) override;
+    String getName() override;
+
+private:
+//    DigitalRainAnimation<TFT_eSPI> matrix_effect = DigitalRainAnimation<TFT_eSPI>();
+    void addConfigToManager();
+    bool m_smallFont = false;
+    int m_textColor; // = 0x001F;
+    int m_headtextColor; // = 0x07FF;
```
flattermann, line 23: I would set some default values here

flattermann (changes requested):
Please also make sure to use the same naming/formatting schema like the rest of the code base:

- **directory** names should be lower case (e.g. `widgets/screensaverwidget` or `widgets/matrixwidget`)
- **classes** and **their files** should be upper/camel case, Widget should be written with a capital W, e.g. `ScreenSaverWidget` or `MatrixWidget`)
- please format the files using the file `.clang-format` in the repo (this will be done automatically if you open the `info-orbs.code-workspace` in VSCode)

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
```diff
@@ -0,0 +1,52 @@
+#include "ScreenSaverwidget.h"
+
+ScreenSaverwidget::ScreenSaverwidget(ScreenManager &manager, ConfigManager &config) : Widget(manager, config){
+    m_enabled = true; 
+    m_config.addConfigBool("ScreenSaverWidget", "SmallFont", &m_smallFont, "Small font (unchecked)", false);
```
flattermann, line 5: Please add a short widget specific prefix to all config keys, e.g. "SmallFont" -> "mtxSmallFont" or similar.

FYI: The key length is limited to 15 characters if I remember correctly.

flattermann: And this should probably be renamed to "Big font".

firmware/config/config.h.template
flattermann: And please do not include it in default build (i.e. comment out the line)

nconner-06: Hi Christian
Many thanks for all your comments and recommendations.
I'll work on updating my little widget and upload revised code as soon as I can.
I'm not an expert on OO coding so might have a few questions for you, if that's ok?
Regards
Neil

flattermann: Hi Neil,
Sure, I'll try my best to answer if I can!

firmware/src/core/screenmanager/ScreenManager.cpp
nconner-06: Hi Christian

I modified the ScreenManager class because I needed access to functions in the base TFT_eSPI class which are not passed by the ScreenManager class, what is the process here?

Can I propose a change to the ScreenManager class in this PR? to make them  available using the standard m_manager object.

May be someone else might need them as some point in the future...
Thanks for the clarification
Neil

flattermann: I see, you need access to `m_tft` to pass it to `DigitalRainAnimation.hpp`.

Well, `.getDisplay()` was public before, but we changed it because a widget should not use it directly.

The best way would probably to use a wrapper call (i.e. a class that implements the `TFT_eSPI` functions used by `DigitalRainAnimation.hpp` and forward them to `ScreenManager`.

I'll do some test as soon as I can.

nconner-06: Ok, thanks
A class within a class and I want to access the internal class...
So, far I have identified I need to access these attributes of TFT_eSPI:
    int16_t  width();
    int16_t  height();
    void     setTextColor(uint16_t c);
    void     setTextColor(uint16_t c, uint16_t b);
    void     setTextColor(uint16_t c, uint16_t b, bool bgfill);
    uint16_t color565(uint8_t r, uint8_t g, uint8_t b);
    void     setCursor(int16_t x, int16_t y);
    void     setTextSize(uint8_t s);
    void     print(String s);
    void     print(char c);

And .getDisplay...

nconner-06: I'm happy to modify the "DigitalRainAnimation.hpp" file if needed.

As you can see the initialization takes a pointer to a TFT_eSPI class object (if I understand the code correctly) 

// initialization
void DigitalRainAnimation::init(TFT_eSPI *gfx,** bool biggerText, bool alphabetOnly)
{
  _gfx = gfx;
  setBigText(biggerText);
  line_len_min = 3;
  line_len_max = 10;
  isAlphabetOnly = alphabetOnly;
  prepareAnim();
}

nconner-06: @flattermann 
FYI, I'm managed to get the function working without needing the protected function.. i.e. I'm using ScreenManager without needing to have any Widget specific code in the core function.
I still do need to add functionality to the SM, but that is just to pass some additional TFT_eSPI functions which are not already presented... 
I'll update this PR during the day

firmware/src/core/screenmanager/ScreenManager.h
nconner-06: ok, removed

firmware/src/main.cpp
nconner-06: ok, removed

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
nconner-06: ok, changed

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
nconner-06: ok, removed

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.h
nconner-06: Defaults are set in the class, but added them here for safety..

firmware/src/widgets/ScreenSaverwidget/ScreenSaverwidget.cpp
nconner-06: good idea, changed.

firmware/config/config.h.template
nconner-06: widget renamed to MatrixWidget
formatting update to follow project standards
Will be disabled as standard and include better comments.

firmware/src/core/screenmanager/ScreenManager.cpp
nconner-06: removed Widget specific reference from the cord code

nconner-06:
I just noticed that it would be good if I add the support for "I18n improvements #241" PR...

Issue #238, opened 2025-02-21
Support translations (i18n)


flattermann:
It would be nice to have support for translations.

I would propose something along these lines:

1. The translation files should be simple key/value pairs, e.g.
```
welcome=Welcome
settings=Settings
language=Language
customclock_x_enabled=CustomClock \1 enabled
```
2. The translation files should be stored in LittleFS
3. The language should be selectable via WebUI
4. On boot, the respective language file will be read as key/value pairs into memory (might that be an issue?)
5. We add a i18n helper class that will lookup the translated string and return the key if it's not found
```
i18n::get("welcome")
```
7. It should support placeholders, e.g.
```
i18n::get("customclock_x_enabled", 5)
```

I'll try to implement a simple proof of concept.

PR #240, opened 2025-02-24
Added Translation support (I18n)

flattermann:
Added support for translations via text files in `/i18n/`.

Each line in the translation files is a simple `key=value` pair.

We parse the files in this order:

1. Load default language file (`en.txt`)
2. Load selected language file (e.g. `de.txt`), use the default language as fallback
3. For every widget, load default language file (`en.widgetname.txt`)
4. For every widget, load selected language file (e.g. `de.widgetname.txt`), use the default language as fallback

We can use `I18n::get("key")` to use the i18n strings.

This also supports parameters, e.g. `I18n::get("version", "2.0")`.

I would recommend for the widget specific files to use a prefix for each key, e.g. `parqet.name=Parqet` to avoid name clashes.

See also #238


flattermann:
I rebased this branch to the current `dev` branch.

The translations work, but there are some limitations with my current approach:

On startup, I'm loading the translation string in a map<String, const char*>. 
That means that the translated strings are stored in RAM. For the strings in this PR, this will consume ~3kB of RAM.
If we convert all WebPortal config strings to i18n, this will probably be around 8-10kB of RAM.

I don't really like that (RAM is precious!), that's why I'd like to try a different approach before merging this PR:

I'll try to put the translations in the source code instead of files on the LittleFS. This will probably allow us to store the translations in PROGMEM instead of RAM. But we lose the ability to download/edit/upload the translations without recompilation.

flattermann:
Ok, I switched to code based translations now.

We have one file for general translations: `src/core/i18n/Translations.cpp` and each widget can also have their own translations, e.g. `src/widgets/parqetwidget/ParqetTranslations.cpp`.

See the two files above for examples.

I've included translations for EN, DE, FR for some WebPortal config options (Sections "General", "TFT Settings" and "ParqetWidget"). 

**FR definitely needs checking, because it's an automated translation.**

PR #241, opened 2025-02-25
I18n improvements

flattermann:
Replaced raw translation array declarations with `Translation` and `TranslationMulti` typedefs to improve readability and maintainability. Updated all translation-related headers and source files to use the new types consistently.


flattermann:
I converted most of the strings of all widgets to i18n.

PR #243, opened 2025-02-27
changes to make it easier to do breakpoint debugging.  Break platform…

dreed47:
changes to make it easier to do breakpoint debugging.  Break platformio.ini environments into a [common] section and a new debug environment.  Also added partitions-debug.csv.  Added comments to template config but also changed TFT_CS to -1 as it needs to be set to something but isnt actually used anywhere and it cant be set to pin 15 or breakpoint debugging fails.


flattermann:
We should really move the defaults away from config.h.

Maybe we can add something like a config.defaults.h with a bunch of 
```
#ifndef XX
  #define XX whatever
#endif
```
This would make the config.h file much cleaner and make it possible to update default values in the future (at the moment, the users would need to copy the new defaults from `config.h.template` to their `config.h`.

We just need to make sure that this defaults file is parsed after `config.h`.

dreed47:
> We should really move the defaults away from config.h.
> 
> Maybe we can add something like a config.defaults.h with a bunch of
> 
> ```
> #ifndef XX
>   #define XX whatever
> #endif
> ```
> 
> This would make the config.h file much cleaner and make it possible to update default values in the future (at the moment, the users would need to copy the new defaults from `config.h.template` to their `config.h`.
> 
> We just need to make sure that this defaults file is parsed after `config.h`.

Or maybe move all non-user variables out of config.h  .. like maybe config.system.h. ??

flattermann:
Yes! That `config.h` looks much cleaner!

platformio.ini
```diff
@@ -9,15 +9,15 @@
 ; https://docs.platformio.org/page/projectconf.html
 
 [platformio]
+default_envs = esp32-v0_3
```
flattermann: Do the names v0_3 and v0_4 refer to the InfoOrbs PCB version?

In that case, I would rename the envs to make this more clear, e.g. infoorbs-v0_3 or similar.

dreed47: yes it really correlates to the PCB version.   Thats a good idea to rename them.

platformio.ini
```diff
@@ -93,3 +92,40 @@ build_flags =
...
+board_build.filesystem = ${common.board_build.filesystem}
+framework = ${common.framework}
+lib_deps = ${common.lib_deps}
+monitor_speed = ${common.monitor_speed}
+extra_scripts = ${common.extra_scripts}
+build_flags = 
+    ${common.build_flags}
+    -D PCB_VERSION="\"infoorbs-v0_3\""
+	-D SCREEN_1_CS=0
+	-D BUTTON_RIGHT_PIN=5
```
flattermann: The non debug environment should probably not override the pins. :-)

dreed47: ops... pushed by mistake

flattermann (approved):

Issue #224, opened 2025-01-12
Output of "garbage" on serial port


lyricnz:
The lines of silly character may be mangled by my terminal, but something is being output there:

```
I: ✅ Released semaphore
I: 🔵 Starting HTTP request for: https://api.twelvedata.com/quote?apikey=e03fc53524454ab8b65d91b23c669cc5&symbol=APC&country=Germany
I: ✅ Released semaphore
����������������������������������������������������������������1
I: Left button short pressed -> switch to prev Widget
displayStock - BTC/USD ~ 93879.93
displayStock - USD/CAD ~ 1.44
displayStock - XEQT ~ 33.56
displayStock - SPY ~ 580.49
displayStock - APC&country=Germany ~ 228.55
Drawing of Stock took 1221 ms
����������������������������������������������������������������
```

dreed47:
I don't see these on my terminal.  Is anyone else seeing them?  I'm on MacOS

dreed47:
I have a theory that these garbage characters are a result of electrical interference.  Let me explain why I say that.  First, I hardly ever see them and I understand many people see them all the time.  This leads me to believe its a local environmental issue.  Second they seem to appear at random points in the app execution.  Third they seem to print without a line feed.  I did a search through the code to find places where we do a Serial.print (not Serial.println) and there are only a few and none of them seem like likely candidates.   Electrical interference can come from bad usb cables, nearby devices, motors, or power supplies. Long cables act as antennas, picking up noise and degrading the signal.  Adjacent wires interfering with each other can corrupt the data.

flattermann:
Startup garbage character should be taken care by #246.

But we cannot really do something about electrical interference. :-)

Feel free to reopen if necessary.

lyricnz:
FYI: che and I spent quite a bit of time yesterday digging into this - observing the patterns, cutting out code and features (all the way to turning off wifi, all the widgets, and even emptying the main loop()), and still having the same issue.  Finally, we connected the device directly to PC, and not via USB hub, and problem went away.

PR #246, opened 2025-03-01
Serial monitor fixes 

dreed47:
some cleanup of serial monitor output.  issue #224


firmware/src/main.cpp
```diff
@@ -46,6 +46,14 @@ void addWidgets() {
 void setup() {
     // Initialize global resources
     initializeGlobalResources();
+
+    // Add a delay to allow the serial interface to initialize
+    delay(4000);
```
learn-more: Please note that this will delay startup by 4 seconds

learn-more (approved):

firmware/src/main.cpp
dreed47: thats obviously true but it seems to take that long for the serial interface to init.  Are you asking that I add to the comment?

learn-more: > thats obviously true but it seems to take that long for the serial interface to init. Are you asking that I add to the comment?

No, more or less commenting it for the people that would look at this PR and not understand that.

firmware/config/config.h.template
```diff
@@ -75,6 +75,9 @@
 // LOG_LEVEL_WARNING, LOG_LEVEL_NOTICE, LOG_LEVEL_VERBOSE
 #define LOG_LEVEL LOG_LEVEL_VERBOSE
 
+// Enable if you're having issues with the app outputting garbage characters on startup
+// #define SERIAL_INTERFACE_INIT_DELAY
```
flattermann: Maybe move the 4000 as value here, this would allow users to have lower/higher delays depending on there individual setup.

firmware/src/main.cpp
```diff
@@ -46,9 +46,20 @@ void addWidgets() {
...
     initializeGlobalResources();
+
+#ifdef SERIAL_INTERFACE_INIT_DELAY
+    // Add a delay to allow the serial interface to initialize
+    delay(4000);
+#endif
+
+    // Clear the serial buffer of any garbage
+    while (Serial.available() > 0) {
+        Serial.read();
```
flattermann, line 59: Shouldn't this be moved after `Serial.begin()`?

dreed47: yes I think so, let me move and retest

flattermann (approved):

brett-dot-tech:
@dreed47 a few small conflicts here as a result of pervious merges (I believe this was ches i18n changes)

Issue #171, opened 2024-12-10
Async calls resulting in widget "loading" on boot to be bypassed


brett-dot-tech:
With @dreed47 s #165 fix to bring stocks async, it has resulted in the "loading" boot screen seeing the stocks widget/API calls as having instantly loaded, therefore not showing the loading screen.

This results in the user being able to scroll through widgets prior to the first API calls(s) being done for stocks and can leave the user looking at blank orbs for some time as the stock widget is rendering. 

A proposed solution was moving the "loading" logic to the widgets instead of on boot, essentially having "loading data..."[or similar] be the default state of a widgets orb on boot if instead of just a black screen. 

I like this personally as it also means users can boot to the clock immediately without waiting, which is what I think the majority of folks use it for/leave it on.

flattermann:
Fixed by #247

PR #247, opened 2025-03-01
Add loading data message to stock widget

dreed47:
On initial load show loading data message so user doesn't have to look at blank displays per issue #171

### DIFF
--- a/firmware/src/widgets/stockwidget/StockWidget.cpp
+++ b/firmware/src/widgets/stockwidget/StockWidget.cpp
@@ -38,7 +38,7 @@ void StockWidget::setup() {
 void StockWidget::draw(bool force) {
     m_manager.setFont(DEFAULT_FONT);
     for (int8_t i = 0; i < m_stockCount; i++) {
-        if (!m_initialized) {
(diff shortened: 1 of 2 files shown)


firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -38,11 +38,17 @@ void StockWidget::setup() {
 void StockWidget::draw(bool force) {
     m_manager.setFont(DEFAULT_FONT);
     for (int8_t i = 0; i < m_stockCount; i++) {
-        if (m_stocks[i].isChanged() || force) {
+        if (!m_initialized) {
```
flattermann: This will always show the message on the first `draw()` of the `StockWidget`, no matter if the data was already loaded or not. 

If you change this to
`if (!m_initialized && !m_stocks[i].getSymbol().isEmpty() && m_stocks[i].getTicker().isEmpty()) {`
it should only show the message if the stock is defined (symbol) but not loaded (ticker) yet.

The only problem that would still exist is that if you quickly switch widgets and come back to StockWidget a second time (while the data is still not loaded), it will not show the loading message again. But that's probably not really an issue.

dreed47: good idea, let me test this

flattermann (approved):

Issue #188, opened 2024-12-17
Enhancement - Set hostname  so it can be found on the network


dreed47:
Set hostname to InfoOrbs-<last 6 digits of MAC address> so it can be found on the network

flattermann:
Fixed by #248

PR #248, opened 2025-03-02
Set hostname to InfoOrbs-<last 6 digits of MAC address> so it can be …

dreed47:
…found on the network RE issue   #188

### DIFF
--- a/firmware/src/widgets/wifiwidget/WifiWidget.cpp
+++ b/firmware/src/widgets/wifiwidget/WifiWidget.cpp
@@ -60,6 +60,11 @@ void WifiWidget::setup() {
     m_wifiManager.setCleanConnect(true);
     m_wifiManager.setConnectRetries(5);
 
+    // Set hostname to InfoOrbs-<last 6 digits of MAC address> so it can be found on the network
+    String mac = WiFi.macAddress();
+    mac.replace(":", "");
+    m_wifiManager.setHostname("InfoOrbs-" + mac.substring(mac.length() - 6));
+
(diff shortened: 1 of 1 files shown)


firmware/src/widgets/wifiwidget/WifiWidget.cpp
```diff
@@ -60,6 +60,11 @@ void WifiWidget::setup() {
     m_wifiManager.setCleanConnect(true);
     m_wifiManager.setConnectRetries(5);
 
+    // Set hostname to InfoOrbs-<last 6 digits of MAC address> so it can be found on the network
+    String mac = WiFi.macAddress();
+    mac.replace(":", "");
+    m_wifiManager.setHostname("InfoOrbs-" + mac.substring(mac.length() - 6));
```
flattermann: Maybe it would be a good idea to print the new hostname to Serial here.

Please also set the same name in L104/105 for mDNS.
You can use `m_wifiManager.getWiFiHostname()` there after setting the hostname here.

dreed47: let me give this a shot and restest

dreed47: so .... String mDNSname = m_wifiManager.getWiFiHostname(); ?

I've never really messes with the mDNS stuff

dreed47: updated

flattermann (approved):

lyricnz (approved):

Issue #249, opened 2025-03-02
ClockWidget update is running every loop instead of once a second


dreed47:
The update function for the ClockWidget is running every time it's called from the main loop instead of once a second.  related issue #186 

The main loop calls this update method about every ~5/10ms and the update function is running every time because the m_secondTimerPrev variable is never being updated.

flattermann:
Fixed by #250

Issue #186, opened 2024-12-16
clockwidget update function strange code in first line


HugoBalder:
The if line should be removed, because it is doing nothing useful. The variable m_secondTimerPrev is set only ONCE in the clockwidget.h to 0.
After that it is never assigned a new value.

So this condition is always false, as millis - 0 is always greater than 2000.

void ClockWidget::update(bool force) {
    if (millis() - m_secondTimerPrev < m_secondTimer && !force) {
        return;
    }

flattermann:
Fixed by #250

PR #250, opened 2025-03-02
fix ClockWidget update to run once a second

dreed47:
related issues #249 and #186


firmware/src/widgets/clockwidget/ClockWidget.cpp
```diff
@@ -165,6 +165,8 @@ void ClockWidget::update(bool force) {
 
         m_lastMinuteSingle = m_minuteSingle;
     }
+
+    m_secondTimerPrev = millis();
```
flattermann: I would move this to the top of the function (after the first `if`) to avoid skipped seconds.
This would probably not happen very often, but putting it to the top should not have negative side effects.

flattermann (approved):

lyricnz (approved):

PR #257, opened 2025-03-03
Move the weather API URL to config file

nconner-06:
Move the weather API URL to config.system.h including #ifndef check

### DIFF
--- a/firmware/config/config.system.h
+++ b/firmware/config/config.system.h
@@ -79,6 +79,10 @@
 #define TIMEZONE_API_URL "http://api.timezonedb.com/v2.1/get-time-zone"
 #define WEATHER_API_KEY "XW2RDGD6XK432AF25BNK2A3C7"
 
+#ifndef WEATHER_API_URL
+	#define WEATHER_API_URL "https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/"
+#endif
+
 #define MAX_RETRIES 3
 
 #endif
--- a/firmware/src/widgets/weatherwidget/WeatherWidget.cpp
+++ b/firmware/src/widgets/weatherwidget/WeatherWidget.cpp
@@ -91,7 +91,7 @@ void WeatherWidget::update(bool force) {
 
 bool WeatherWidget::getWeatherData() {
     String weatherUnits = m_weatherUnits == 0 ? "metric" : "us";
-    String httpRequestAddress = "https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/" +
+    String httpRequestAddress = String(WEATHER_API_URL) +
                                 String(m_weatherLocation.c_str()) + "/next3days?key=" + weatherApiKey + "&unitGroup=" + weatherUnits +
                                 "&include=days,current&iconSet=icons1&lang=" + LOC_LANG;
 


dreed47 (commented):
I agree this is a needed change.  Ideally the end user should be able to choose their weather provider and configure the needed variables.  One issue will be making the parsing code specific to the weather API thats chosen

flattermann (approved):

nconner-06:
@dreed47, 
Which weather providers do we want to support? Visual Crossing and OpenWeatherMap (at least), who else?
Neil

dreed47:
> @dreed47, Which weather providers do we want to support? Visual Crossing and OpenWeatherMap (at least), who else? Neil

I would like to have Tempest weather for people that have their own Tempest weather station.  But also ...

I had a working approach a while back that broke out the weather feed into separate feeds like /feeds/VisualCrossing.cpp.  Then it conditionally added the correct feed with something like.

```
WeatherFeed *WeatherWidget::createWeatherFeed() {
#ifdef WEATHER_TEMPEST_FEED
    return new TempestFeed(WEATHER_TEMPEST_API_KEY, WEATHER_TEMPEST_STATION_ID, m_weatherUnits, WEATHER_TEMPEST_STATION_NAME);
#else
    return new VisualCrossingFeed(WEATHER_VISUALCROSSING_API_KEY, m_weatherLocation);
#endif
}
``` 

It may take a lot of work to update it but im looking for the code right now.

nconner-06:
it might be an idea to switch to Lat/lon coordinates and not place names for the location.  I know that OpenWeatherMap are stopping their Geolocalisation - in fact the new API 3.0 only works with Lat/Lon coordinates.  Which is why I changed it on the version I did, using the 'place name' just as a user defined display name - so can be anything, like "home", etc..

dreed47:
> it might be an idea to switch to Lat/lon coordinates and not place names for the location. I know that OpenWeatherMap are stopping their Geolocalisation - in fact the new API 3.0 only works with Lat/Lon coordinates. Which is why I changed it on the version I did, using the 'place name' just as a user defined display name - so can be anything, like "home", etc..

Im not sure how far you are with your version but I have mine working for VisualCrossing as a separate Feed class in a /feeds/ subfolder.  I'm working to get my old Tempest weather code implemented next.  If you want I can push my working branch to my repo if you want to take a look to see how it compares with yours?.

nconner-06:
I didn't split the feeds out to separate files, just added a build "#ifdef" section around the code in the WeatherWidget code...
I had a look at your branch "weather-plugable" and saw what you had done - is this the latest version?
I don't have an account for "Tempest", so haven't tried them to see if they have good coverage.  The VisualCrossing doesn't have good coverage in my region, which is why I looked at OpenWeatherMap.

dreed47:
> I didn't split the feeds out to separate files, just added a build "#ifdef" section around the code in the WeatherWidget code... I had a look at your branch "weather-plugable" and saw what you had done - is this the latest version? I don't have an account for "Tempest", so haven't tried them to see if they have good coverage. The VisualCrossing doesn't have good coverage in my region, which is why I looked at OpenWeatherMap.
No , thats the old branch...look at branch "weatherwidget-feeds" that I just pushed.  It should compile and run the visual crossing weather okay.   Tempest is a speacial use case as its primarily for ppl that have their own Tempest weather station.  I'm not sure if they have a way to get a key without having a station.

nconner-06:
This is now handled via another PR, so closing it.

PR #258, opened 2025-03-03
Convert StockWidget to ArduinoLog per #220

dreed47:
Convert stockwidget arduino log

### DIFF
--- a/firmware/src/widgets/stockwidget/StockWidget.cpp
+++ b/firmware/src/widgets/stockwidget/StockWidget.cpp
@@ -108,7 +108,7 @@ void StockWidget::buttonPressed(uint8_t buttonId, ButtonState state) {
 }
 
 void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint32_t backgroundColor, uint32_t textColor) {
-    Log.infoln("displayStock - %s ~ %d", stock.getSymbol().c_str(), stock.getCurrentPrice());
+    Log.infoln("displayStock - %s ~ %s", stock.getSymbol().c_str(), stock.getCurrentPrice(2).c_str());
     if (stock.getCurrentPrice() == 0.0) {
         // There isn't any data to display yet
         return;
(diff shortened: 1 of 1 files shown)


firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -87,13 +88,13 @@ void StockWidget::processResponse(StockDataModel &stock, int httpCode, const Str
                 stock.setTicker(doc["symbol"].as<String>());
                 stock.setCurrencySymbol(doc["currency"].as<String>());
             } else {
-                Serial.println("skipping invalid data for: " + stock.getSymbol());
+                Log.warningln("skipping invalid data for: %s", stock.getSymbol());
```
flattermann: I think we still need .c_str() for Arduino Strings, right?

dreed47: yep

firmware/src/widgets/stockwidget/StockWidget.cpp
```diff
@@ -107,7 +108,7 @@ void StockWidget::buttonPressed(uint8_t buttonId, ButtonState state) {
 }
 
 void StockWidget::displayStock(int8_t displayIndex, StockDataModel &stock, uint32_t backgroundColor, uint32_t textColor) {
-    Serial.println("displayStock - " + stock.getSymbol() + " ~ " + stock.getCurrentPrice());
+    Log.infoln("displayStock - %s ~ %d", stock.getSymbol(), stock.getCurrentPrice());
```
flattermann: One more `.c_str()`, please. :-)

dreed47: yep

lyricnz: `stock.getCurrentPrice()` returns a float, not a decimal - so you probably need a `%9.6f` type format or pass a "digits" so you get a string

PR #259, opened 2025-03-03
Proposed Developer logging guide

dreed47:
feedback would be great!

### DIFF
--- a/references/Developer Guide to Logging.md
+++ b/references/Developer Guide to Logging.md
@@ -27,7 +27,7 @@ When using the `ArduinoLog` library, the logging methods correspond to different
    - Corresponds to `Log.warning()`.
    - Example: `Log.warning("This is a warning");`
 
-5. **`LOG_LEVEL_INFO` and `LOG_LEVEL_NOTICE` **:
+5. **`LOG_LEVEL_INFO` and `LOG_LEVEL_NOTICE`**:
 
    - Informational messages, warnings, and errors.
    - Corresponds to `Log.info()` and `Log.notice()`.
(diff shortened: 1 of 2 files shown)


lyricnz:
Looks good! I started raising pr to fix the code. Will check it meets your standards. Was avoiding mega pr since it gets stale super quick and I wasn’t sure I had consensus

flattermann:
What about the ...ln (newline) functions, e.g. `Log.warningln()`.
Should these be mentioned in the guide?
Or do we always print a newline no matter what?

Also, it seems that INFO is replaced by NOTICE (not the other way around), so we should probably encourage using NOTICE as well:
https://github.com/thijse/Arduino-Log/blob/master/README.md#log-events

references/Developer Guide to Logging.md
```diff
@@ -0,0 +1,87 @@
...
+   - Corresponds to `Log.error()`.
+   - Example: `Log.error("This is an error");`
+
+4. **`LOG_LEVEL_WARNING`**:
+
+   - Warnings and errors.
+   - Corresponds to `Log.warning()`.
+   - Example: `Log.warning("This is a warning");`
+
+5. **`LOG_LEVEL_INFO` and `LOG_LEVEL_NOTICE` **:
```
lyricnz: Formatting seems a little broken

image

lyricnz:
> What about the ...ln (newline) functions, e.g. `Log.warningln()`. Should these be mentioned in the guide? Or do we always print a newline no matter what?

Agree - it's worth a note. They appear to be identical under the covers, which makes me think the shorter version is better.


```cpp
  template <class T, typename... Args> void notice(T msg, Args...args){
#ifndef DISABLE_LOGGING
    printLevel(LOG_LEVEL_NOTICE, false, msg, args...);
#endif
  }
  
  template <class T, typename... Args> void noticeln(T msg, Args...args){
#ifndef DISABLE_LOGGING
    printLevel(LOG_LEVEL_NOTICE, true, msg, args...);
#endif
  }  

  template <class T, typename... Args> void info(T msg, Args...args) {
#ifndef DISABLE_LOGGING
	  printLevel(LOG_LEVEL_INFO, false, msg, args...);
#endif
  }

  template <class T, typename... Args> void infoln(T msg, Args...args) {
#ifndef DISABLE_LOGGING
	  printLevel(LOG_LEVEL_INFO, true, msg, args...);
#endif
  }
```

flattermann:
The difference is that ...ln() adds a CR: :-)

```
  template <class T, typename... Args> void notice(T msg, Args...args){
#ifndef DISABLE_LOGGING
    printLevel(LOG_LEVEL_NOTICE, **false**, msg, args...);
#endif
  }
  
  template <class T, typename... Args> void noticeln(T msg, Args...args){
#ifndef DISABLE_LOGGING
    printLevel(LOG_LEVEL_NOTICE, **true**, msg, args...);
#endif
  }  
```

```
template <class T> void printLevel(int level, bool cr, T msg, ...)
	{
#ifndef DISABLE_LOGGING
...
		if (cr)
		{
		    _logOutput->print(CR);
		}
#endif
	}
```

lyricnz:
Ahh!  I missed that bool, was looking at `msg`!

flattermann:
The guide should also mention to not pass Arduino Strings directly for `%s`, but to use `.c_str()`, like in `Serial.printf()`. 
Otherwise we might get garbage characters, I think.

dreed47:
> What about the ...ln (newline) functions, e.g. `Log.warningln()`. Should these be mentioned in the guide? Or do we always print a newline no matter what?
> 
> Also, it seems that INFO is replaced by NOTICE (not the other way around), so we should probably encourage using NOTICE as well: https://github.com/thijse/Arduino-Log/blob/master/README.md#log-events

There appears to be conflicts within their docs and code base regarding INFO and NOTICE.  Their header file says this 

` * 0 - LOG_LEVEL_SILENT     no output
 * 1 - LOG_LEVEL_FATAL      fatal errors
 * 2 - LOG_LEVEL_ERROR      all errors
 * 3 - LOG_LEVEL_WARNING    errors and warnings
 * 4 - LOG_LEVEL_INFO       errors, warnings and notices
 * 4 - LOG_LEVEL_NOTICE     Same as INFO, kept for backward compatibility
 * 5 - LOG_LEVEL_TRACE      errors, warnings, notices, traces
 * 6 - `

As far as I can tell it treats them both the same as level #4. We can probably just decide which we prefer ppl to use and recommend that in the docs

flattermann:
It should be `...ln()` not `...nl()`. :-)

lyricnz (approved):

PR #260, opened 2025-03-03
Convert MQTTWidget to use ArduinoLog per #220

dreed47:
replaced Serial prints with appropriate Log methods


flattermann:
All the `%s` here should have a `.c_str()` as well.

And yes, that was missing all along. :-)

Issue #252, opened 2025-03-03
Add (optional?) timestamp to serial output


lyricnz:
This would help us identify the timing when working on issues that may be related (eg: network or NTP or background updates etc). Relates to #220

lyricnz:
Right now, output is a mixture of `Log` output and `Serial.print`.  Making output a bit of a mixed bag:

<img width="424" alt="Image" src="https://github.com/user-attachments/assets/175aa412-0325-4601-acf1-9afbf6fbf5e8" />

lyricnz:
Q: should the logging use the native time on the device (which appears to start at 00:00:00 - see above), or the time we get from NTP (GlobalTime)?

flattermann:
Implemented in #261

lyricnz:
I'm going to add support for printing the NTP time, not just the hardware/offset time.

PR #261, opened 2025-03-03
Include timestamp (offset) in log messages

lyricnz:
Per https://github.com/brettdottech/info-orbs/issues/252

### DIFF
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -75,6 +75,9 @@
 // LOG_LEVEL_WARNING, LOG_LEVEL_NOTICE, LOG_LEVEL_VERBOSE
 #define LOG_LEVEL LOG_LEVEL_VERBOSE
 
+// Include current time in all log output
+// #define LOG_TIMESTAMP
+
 // Enable if you're having issues with the app outputting garbage characters on startup, delay in ms
 // #define SERIAL_INTERFACE_INIT_DELAY 4000
 
--- a/firmware/src/core/utils/MainHelper.cpp
+++ b/firmware/src/core/utils/MainHelper.cpp
@@ -5,6 +5,7 @@
 #include "icons.h"
 #include <ArduinoLog.h>
 #include <HTTPClient.h>
+#include <TimeLib.h>
 #include <esp_task_wdt.h>
 
 static Button buttonLeft;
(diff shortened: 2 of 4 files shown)


dreed47:
I think this is fine and much simpler than what I have.  Can we wrap it in an #ifdef so its off by default?

flattermann:
I would move `printPrefix()` to MainHelper to unclutter main.cpp and I would keep it enabled all the time (or is there a reason anyone would want to disable it?).

Also, after merging #262, we could actually use `GlobalTime::getInstance()->isTimeSet()` to switch between `millis()` and real time.

dreed47:
> I would move `printPrefix()` to MainHelper to unclutter main.cpp and I would keep it enabled all the time (or is there a reason anyone would want to disable it?).
> 
> Also, after merging #262, we could actually use `GlobalTime::getInstance()->isTimeSet()` to switch between `millis()` and real time.

my preference would be to not have it on all the time but I won't lose any sleep over it if it is .

dreed47:
> I would move `printPrefix()` to MainHelper to unclutter main.cpp and I would keep it enabled all the time (or is there a reason anyone would want to disable it?).
> 
> Also, after merging #262, we could actually use `GlobalTime::getInstance()->isTimeSet()` to switch between `millis()` and real time.

logging is initialized before we have NTP correct?  You would have to account for that in the printPrefix wrapper

flattermann:
Yes, and until NTP is ready, `GlobalTime::getInstance()->isTimeSet()` should be false.

lyricnz:
Moved time-logging to MainHelper and make it #ifdef off by default

flattermann (approved):

lyricnz:
@flattermann there doesn't seem to be any `GlobalTime::getInstance()->isTimeSet()`.  How can we detect if there is a valid time?   (m_unixEpoch is not initialized, so IIRC it's undefined until `::updateTime()` has been called successfully)

I'd also prefer not to *cause* instantiation of GlobalTime when checking this in the log-prefix (otherwise we run the risk of a recursive loop if that calls logging).   Maybe adding a static method to GlobalTime?

```cpp
time_t GlobalTime::getUnixEpochIfAvailable() {
    return m_instance ? m_instance->getUnixEpoch() : 0;
}
```

as well as initializing `m_unixEpoch=0` in the .h file?

lyricnz:
@flattermann https://github.com/brettdottech/info-orbs/pull/264 ??

flattermann:
@lyricnz Yes, you are right, we would need access to `m_timeClient->isTimeSet()`.

`GlobalTime::getUnixEpochIfAvailable()` sounds like a good idea.

PR #262, opened 2025-03-04
Integrate NTPClient library to fix upstream bug

flattermann:
Integrate NTPClient library to fix upstream bug. 
This should allow us to simplify the logic in GlobalTime.cpp. 
The clock is kept at 0:00 until we got a valid NTP response.

See https://github.com/arduino-libraries/NTPClient/pull/211/files

### DIFF
--- a/.mega-linter.yml
+++ b/.mega-linter.yml
@@ -28,4 +28,4 @@ FILEIO_REPORTER: false
 # Uncomment if you want MegaLinter to detect errors but not block CI to pass
 # DISABLE_ERRORS: true
 
-CPP_CLANG_FORMAT_FILTER_REGEX_EXCLUDE: firmware/lib/OpenFontRender/.*
+CPP_CLANG_FORMAT_FILTER_REGEX_EXCLUDE: firmware/lib/.*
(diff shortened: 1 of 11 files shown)


lyricnz (approved):
Can we add a commit message describing the source of the lib: a link to the upstream repo, plus the commit hash, plus the PR applied to fix the bug.

flattermann:
@lyricnz Like this?

lyricnz:
"this"?  Sorry, I don't see it?   I would think something like:

> This library is a fork of https://github.com/arduino-libraries/NTPClient <master> at 61684d4 with the patch from https://github.com/arduino-libraries/NTPClient/pull/211/files

flattermann:
@lyricnz Take a look at the individual commits:
https://github.com/brettdottech/info-orbs/pull/262/commits/b515f68dda603c5f8950f52fa5e87e6510902f39
https://github.com/brettdottech/info-orbs/pull/262/commits/a2f3127e239915fadb7c98282b77b6144059144e

lyricnz:
Thanks!

lyricnz:
Yay!

Issue #227, opened 2025-01-13
Invalid time on startup until NTP query succeeds


lyricnz:
On startup, the m_unixEpoch is set to some strange value (-28788 here) which persists until the clock updates via NTP some time later.  If we don't have a real (reliable) value, we should use a zero/undefined, and not use it to display widgets etc.  Related to #226 and #225 

```
Connected to WiFi
updating widget Clock
updating widget Weather
updating widget Stock
GlobalTime::updateTime - inside1: m_timeZoneOffset: -1; m_nextTimeZoneUpdate=0; m_unixEpoch=0
GlobalTime::updateTime - inside2
Timezone Offset from API: -28800
Next timezone update: 1741514528
I: 🔵 Starting HTTP request for: https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/Victoria, BC/next3days?key=XW2RDGD6XK432AF25BNK2A3C7&unitGroup=metric&include=days,current&iconSet=icons1&lang=en
GlobalTime::updateTime - inside1: m_timeZoneOffset: -28800; m_nextTimeZoneUpdate=1741514528; m_unixEpoch=-28788
GlobalTime::updateTime - inside2
Timezone Offset from API: -28800
Next timezone update: 1741514656
I: ✅ Released semaphore
...
Next timezone update: 1741514540
GlobalTime::updateTime - inside1: m_timeZoneOffset: -28800; m_nextTimeZoneUpdate=1741514540; m_unixEpoch=-28703
GlobalTime::updateTime - inside2
Timezone Offset from API: -28800
Next timezone update: 1741514563
GlobalTime::updateTime - inside3 - got result 1736780184
GlobalTime::updateTime - inside1: m_timeZoneOffset: -28800; m_nextTimeZoneUpdate=1741514563; m_unixEpoch=1736780184
GlobalTime::updateTime - inside1: m_timeZoneOffset: -28800; m_nextTimeZoneUpdate=1741514563; m_unixEpoch=1736780185
GlobalTime::updateTime - inside1: m_timeZoneOffset: -28800; m_nextTimeZoneUpdate=1741514563; m_unixEpoch=1736780186
```

flattermann:
#245 ignores invalid dates (before 2025 and after 2035).

lyricnz:
Will test this again tomorrow. We got distracted chasing the "noise" on the serial terminal

flattermann:
[...]

PR #264, opened 2025-03-04
Show NTP time in logs (if available)

lyricnz:
See #261 and #227

### DIFF
--- a/firmware/src/core/utils/MainHelper.cpp
+++ b/firmware/src/core/utils/MainHelper.cpp
@@ -517,7 +517,7 @@ void MainHelper::printPrefix(Print *_logOutput, int logLevel) {
     unsigned long now_ms = millis();
     time_t now_s = GlobalTime::getUnixEpochIfAvailable(); // local time or zero
     if (now_s == 0) {
-        now_s = now_ms % 1000; // Fall back to hardware time if we don't have a valid time yet
+        now_s = now_ms / 1000; // Fall back to hardware time if we don't have a valid time yet
     }
     char timestamp[20];
     sprintf(timestamp, "%02d:%02d:%02d.%03d ", hour(now_s), minute(now_s), second(now_s), now_ms % 1000);
--- a/firmware/src/core/utils/MainHelper.cpp
+++ b/firmware/src/core/utils/MainHelper.cpp
@@ -514,12 +514,18 @@ void MainHelper::watchdogReset() {
 }
 
 void MainHelper::printPrefix(Print *_logOutput, int logLevel) {
-    unsigned long now_ms = millis();
-    time_t now_s = GlobalTime::getUnixEpochIfAvailable(); // local time or zero
-    if (now_s == 0) {
-        now_s = now_ms / 1000; // Fall back to hardware time if we don't have a valid time yet
-    }
     char timestamp[20];
-    sprintf(timestamp, "%02d:%02d:%02d.%03d ", hour(now_s), minute(now_s), second(now_s), now_ms % 1000);
+#ifdef LOG_NTP_TIME
+    time_t now_s = GlobalTime::getUnixEpochIfAvailable(); // local time or zero
+#else
+    time_t now_s = 0;
+#endif
+    if (now_s == 0) {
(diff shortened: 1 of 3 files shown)


lyricnz:
This is how the switchover looks:

image

firmware/src/core/utils/MainHelper.cpp
```diff
@@ -514,7 +515,10 @@ void MainHelper::watchdogReset() {
 
 void MainHelper::printPrefix(Print *_logOutput, int logLevel) {
     unsigned long now_ms = millis();
-    time_t now_s = now_ms / 1000;
+    time_t now_s = GlobalTime::getUnixEpochIfAvailable(); // local time or zero
+    if (now_s == 0) {
+        now_s = now_ms % 1000; // Fall back to hardware time if we don't have a valid time yet
```
flattermann: This should probably not be the modulo operator, right?

flattermann: Also, what happens on `now_s = millis()/1000` and we use `minute(now_s)`, `second(now_s)` etc. Is that correct?
If not, we could use this conversion:
https://github.com/thijse/Arduino-Log/blob/3f4fcf5a345c1d542e56b88c0ffcb2bd2a5b0bd0/examples/Log-advanced/Log-advanced.ino#L52

flattermann: `now_s = millis()/1000` ?

lyricnz: Sorry, I don't understand your question?   After the code below:
```cpp
    time_t now_s = GlobalTime::getUnixEpochIfAvailable(); // local time or zero
    if (now_s == 0) {
        now_s = now_ms / 1000; // Fall back to hardware time if we don't have a valid time yet
    }
```

now_s now contains a UnixEpoch (in seconds), which is either the one from NTP, or the hardware one, which is seconds since power-on.  So when we emit it:

```cpp
sprintf(timestamp, "%02d:%02d:%02d.%03d ", hour(now_s), minute(now_s), second(now_s), now_ms % 1000);
```

That either emits a real local time (if available), or otherwise the hardware time (offset).  You can even see it switch in the logging below, once the NTP call works and we have a valid GlobalTime:

```
00:00:11.761 I: TaskParams created: 2
00:00:11.763 I: TaskParams created: 3
00:00:11.764 I: TaskParams created: 4
00:00:11.765 I: TaskParams created: 5
00:00:11.766 I: TaskParams created: 6
Timezone Offset from API: 39600
Next timezone update: 1743868958
12:30:08.800 I: ✅ Obtained semaphore
12:30:08.801 I: Active requests: 1 (Max seen: 1)
12:30:08.802 I: Processing request: https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/Somerville, Victoria, Australia/next3days?key=REMOVED&unitGroup=metric&include=days,current&iconSet=icons1&lang=en (Remaining in queue: 5)
12:30:08.823 I: 🔵 Starting HTTP request for: https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/Somerville, Victoria, Australia/next3days?key=REMOVED&unitGroup=metric&include=days,current&iconSet=icons1&lang=en
12:30:11.538 I: Active requests now: 0
```

lyricnz: PS: those `hour(time_t t)` functions are just things that parse the `t` into human date/time, and return bits of it.

lyricnz: The reason I didn't inline the call to `millis()` was that we also use it to add the ms on the timestamp.  Which is, to be fair, probably not accurate for NTP time (our library only returns second-accuracy).

Should I make two different printf() - one with ms accuracy (for hardware time) and one with s accuracy (for NTP time)?  I might add another ifdef to turn off NTP time completely, since it's semi-useless for me without ms accuracy.

Alternatively, we could apply another of the unmerged patches in NTPClient which adds ms accuracy....

lyricnz: Pushed a change to make NTP/hardware time configurable, and only show ms when using hardware.

PR #266, opened 2025-03-05
Make WeatherWidget data feeds pluggable and add support for  Tempest and OpenWeatherMap Weather feeds

dreed47:
- Moved VisualCrossing feed functionality info a feeds/VisualCrossingFeed.cpp
- Added a VisualCrossingFeed::setupConfig() method so feeds can add their feed specific config fields
- Added support for a Tempest Weather feed.  NOTE: currently using a temporary proxy hosted by me, need a long term solution for proxys in general.  [proxy code](https://github.com/dreed47/info-orbs-tempest-proxy) (similar to ParqetWidget)
- Added support for OpenWeatherMap from @nconner-06
- If #define WEATHER_OPENWEATHERMAP_FEED or WEATHER_TEMPEST_FEED is not define then the default VisualCrossing weather API is used
- Made changes to TaskFactory to support non ssl requests if the url is http instead of https, this is useful for running local proxys that may not use ssl certs


nconner-06:
I also have the code for a feed from "PirateWeather" using this PR, so once it's included in the DEV branch, I can create a second PR.

PR #267, opened 2025-03-06
Use newer esptool

HeMan:
With updated ESPTool you don't need to push any button to flash over cable


dreed47 (changes requested):
I don't know what this package does but common settings should be done in the [common] section so they don't have to be repeated in all the environment sections

HeMan:
> I don't know what this package does but common settings should be done in the [common] section so they don't have to be repeated in all the environment sections

Should I move `platform` and `board` as well while I'm at it?

PR #269, opened 2025-03-06
Switch to new API provider for the timeZone information

nconner-06:
Objective of the change:
- Using the "TIMEZONE_API" variable in config.system.h to switch between API providers for time zone information.

Benefits:
- The new API doesn't require an API key
- The new API doesn't have a limit of 1 call per second (which is a feature required for a multi-zone clock).

I have made the modification backwards compatible with the current version for calculating when the daylight savings change will occur.

### DIFF
--- a/firmware/config/config.system.h
+++ b/firmware/config/config.system.h
@@ -102,7 +102,7 @@
 #endif
 
 #ifndef TIMEZONE_API_URL
-    #define TIMEZONE_API_URL "https://timeapi.io/api/timezone/zone"
+    #define TIMEZONE_API_URL "http://timeapi.io/api/timezone/zone"
 #endif
 
 #ifndef WEATHER_API_KEY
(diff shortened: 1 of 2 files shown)


brett-dot-tech:
@nconner-06 - a few small conflicts here from past merges sorry. This looks like a much better API than the one I'm paying for. I am fine to switch it out entirely if you see fit, I dont see any downsides personally.

If you want to make those tweaks I'll get those merged asap

brett-dot-tech:
@nconner-06 as per our conversation in disc, why dont we go ahead and move over entirely to this new API and sunset the old one, it seems to do a better job in ever aspect

brett-dot-tech:
@nconner-06 for some reason I am not getting a boot loop when attempting to flasg this, devices resets every ~5 seconds while attempting to connect to wifi.

This behaviour is not shown on other branches for me

brett-dot-tech:
@nconner-06 actually disregard this, it _may_ Be a hardware issue on my end and I will need to test further when I arrive home

PR #270, opened 2025-03-06
Update weather widget logic

dgutzmann:
# Description
Small tweaks and improvements to the weather widget

## Changes
* Updates to reference the config as member config
* Consolidated the `changeMode` logic
 *  Moving the `m_prevMillisSwitch` to keep the timer from switching right after a "manual" switch via button press


brett-dot-tech:
These are great thank you (:

PR #273, opened 2025-03-09
Switching to using extends

dgutzmann:
# Description
Switching to using the `extends` keyword instead of duplicating the config.  [reference](https://docs.platformio.org/en/latest/projectconf/sections/env/options/advanced/extends.html)


dreed47:
these changes are covered in PR #319

Issue #277, opened 2025-03-13
Remove hard dependency on config.h


dreed47:
Since most of the configuration that a typical user needs can be done through the web interface, we should not impose a hard dependency on copying the config.h.template file to config.h for those who choose to manually compile and deploy the code. This is a non-issue for users utilizing the web flasher. However, for those not using the web flasher, it should be possible to clone the repository, compile the code (with reasonable defaults), deploy it, and then configure Wi-Fi and other settings through the web interface.

That said, the ability to use a config.h file should still be available—it just shouldn’t be mandatory.

dreed47:
Meged

PR #278, opened 2025-03-13
Remove hard dependency on config.h #277

dreed47:
- Move config.h variables to config.system.h with defaults.
- Modify python scripts to also look at config.system.h as well as config.h.  
- Modify main.cpp to check if widget includes are not set to WIDGET_DISABLED before including them.
- Restructured config.h.template to better distinguish between web-based settings and non web-based settings
- Changed widget webui enable switch to look at WIDGET_ENABLED 

issue #277


firmware/config/config_helper.h
```diff
@@ -1,6 +1,5 @@
-#if !__has_include("config.h")
-    #error "You need to copy firmware/config/config.h.template to firmware/config/config.h and modify it to match your configuration"
-#else
+// Include config.h if it exists
+#ifdef USE_CONFIG_H
```
learn-more:
```suggestion
#if __has_include("config.h")
```
This should do it automatically?

No need to specify the `USE_CONFIG_H` somewhere.

.gitignore
```diff
@@ -14,7 +14,9 @@
 
 # Arduino libraries (if installed in the project directory)
 libraries/
-firmware/config/config.h
+firmware/config/config.h*
+firmware/config/orb_config.h*
```
learn-more: This also includes the template, which very much should not be under gitignore (you did not commit it here)

firmware/config/config_helper.h
```diff
@@ -1,6 +1,6 @@
 // Include config.h if it exists
```
learn-more, line 1: nit: you did not update the name in this comment

dreed47: fixed

flattermann:
Is it really necessary to rename `config.h` to `orb_config.h`?
I think `config.h` (and `config.system.h`) is a cleaner name and it worked before...

dreed47:
> Is it really necessary to rename `config.h` to `orb_config.h`? I think `config.h` (and `config.system.h`) is a cleaner name and it worked before...

Using __has_include("config.h") in config_helper.h when we have no config.h file throws errors because of the other config.h files in our build.  This is because the compiler searches for config.h in the include paths specified in the build system and we have many libraries that use a config.h the compiler is picking the wrong one giving us linker errors like ```undefined reference to _impure_ptr```.   Generally its a good practice to not use a generically named config.h file but instead one that's unique to the app.  Let me see if I can find a solution for the linker error.  I might also post the question to the core dev group

brett-dot-tech:
This will need a small rebase for the new 5zone clock widget @dreed47, then i'll get er merged. Thanks for the patience there.

Issue #288, opened 2025-03-15
Use a single config for weather feed


flattermann:
Instead of
```
// WEATHER CONFIGURATION (ONLY SET ONE FEED TO TRUE)
#define WEATHER_VISUALCROSSING_FEED true
#define WEATHER_OPENWEATHERMAP_FEED false
#define WEATHER_TEMPEST_FEED false
```

I think an enum would be better, e.g.

```
enum WeatherFeedType {
  VISUALCROSSING = 0,
  OPENWEATHERMAP,
  TEMPEST
}
```

```
// WEATHER CONFIGURATION
#define WEATHER_FEED VISUALCROSSING
```

This would look cleaner in the config.h.

nconner-06:
then change the createWeatherFeed function to a 'switch' statement:
    switch (WEATHER_FEED){
        case (VISUALCROSSING): {
            return new VisualCrossingFeed(WEATHER_VISUALCROSSING_API_KEY, weatherUnits);        
        }
        case (OPENWEATHERMAP): {
            return new OpenWeatherMapFeed(WEATHER_OPENWEATHERMAP_API_KEY, weatherUnits);        
        }
        case (TEMPEST): {
            return new TempestFeed(WEATHER_TEMPEST_API_KEY, weatherUnits);
        }
    }

Issue #287, opened 2025-03-15
Modify WeatherWidget to use float rather than string for the LAT/Long values


nconner-06:
As the PR #281 and PR #266 have been merged, I propose we move the variables for LAT and LOG to float.

Also, move VisualCrossing to use LAT & LONG values rather than place name

I have another feed which also uses LAT/LONG, so that is a better standard

If all feeds (except Tempest) user LAT/LONG, should we just have one field for them or one per feed?

nconner-06:
PR #289 created for this

PR #289, opened 2025-03-15
Dev weather - linked to issue #287 / #298 / #288 plus additional feed

nconner-06:
Move the variables for LAT and LOG to float for OpenWeathermap to floats


nconner-06:
This PR handles the following issues:
- Modify WeatherWidget to use float rather than string for the LAT/Long values #287
- Use a single config for weather feed #288
- use Lat/Long for Visual Crossing weather #298 
- 
And includes a new feed from Pirate Weather

firmware/config/config.h.template
```diff
@@ -35,24 +35,41 @@
 #define DEFAULT_CLOCK ClockType::NORMAL // Set the default clock face: ClockType::NORMAL, ClockType::NIXIE or ClockType::CUSTOM0..9
 //#define CLOCK_FONT DSEG7              // Define clock font (defaults to DSEG7, you can also try DSEG14 or FINAL_FRONTIER)
 
-// WEATHER CONFIGURATION (ONLY SET ONE FEED TO TRUE)
-#define WEATHER_VISUALCROSSING_FEED true
-#define WEATHER_OPENWEATHERMAP_FEED false
-#define WEATHER_TEMPEST_FEED false
+// WEATHER CONFIGURATION
+enum WeatherFeedType {
```
flattermann: The enum should not be defined in `config.h.template`.

Maybe move it to `WeatherWidget.h`?

firmware/config/config.h.template
```diff
@@ -35,24 +35,41 @@
...
 #define WEATHER_VISUALCROSSING_LOCATION "Victoria, BC" // City/state for the weather, look it up on https://www.visualcrossing.com/weather-data
-
-//#define WEATHER_OPENWEATHERMAP_API_KEY "" // Your OpenWeatherMap API key (get it from https://home.openweathermap.org/users/sign_up)
-//#define WEATHER_OPENWEATHERMAP_LAT "" // Latitude of your location
-//#define WEATHER_OPENWEATHERMAP_LON "" // Longitude of your location
-//#define WEATHER_OPENWEATHERMAP_NAME "Victoria" // Name of your location
-
-//#define WEATHER_TEMPEST_STATION_ID "136542" // Set to the station ID of your Tempest weather station, or find one close to you at https://tempestwx.com/map
-//#define WEATHER_TEMPEST_STATION_NAME "Chicago" // Give your Tempest weather station a name
+#define WEATHER_VISUALCROSSING_API_URL "https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/"
```
flattermann: The API URLs should be moved to config.system.h (with ifndef), because a user shouldn't need to override these.

firmware/config/config.h.template
```diff
@@ -35,24 +35,41 @@
...
+#define WEATHER_VISUALCROSSING_API_KEY "XW2RDGD6XK432AF25BNK2A3C7" // Your Visual Crossing API key (https://www.visualcrossing.com/sign-up/)
```
flattermann: The visual crossing API key can also be moved to `config.system.h` (with ifndef, so that it could be overriden here)

firmware/config/config.h.template
```diff
@@ -35,24 +35,41 @@
...
+#define WEATHER_VISUALCROSSING_LAT 0.0 // Latitude of your location
```
flattermann: The three other `WEATHER_VISUALCROSSUNG_*` defs should have a default in `config.system.h` (because we want to be able to build without `config.h`

firmware/config/config.h.template
```diff
@@ -35,24 +35,41 @@
...
-#define WEATHER_OPENWEATHERMAP_FEED false
-#define WEATHER_TEMPEST_FEED false
+// WEATHER CONFIGURATION
+enum WeatherFeedType {
+    VISUALCROSSING = 0,
+    OPENWEATHERMAP = 1,
+    PIRATEWEATHER=2,
+    TEMPEST = 3
+  };
+#define WEATHER_FEED VISUALCROSSING
```
flattermann, line 39: WEATHER_FEED needs a default in `config.system.h` (for `config.h` less builds)

firmware/src/widgets/weatherwidget/WeatherWidget.cpp
```diff
@@ -38,13 +39,20 @@ WeatherFeed *WeatherWidget::createWeatherFeed() {
...
     int weatherUnits = m_config.getConfigInt("weatherUnits", m_weatherUnits);
 
-#if WEATHER_OPENWEATHERMAP_FEED
-    return new OpenWeatherMapFeed(WEATHER_OPENWEATHERMAP_API_KEY, weatherUnits);
-#elif WEATHER_TEMPEST_FEED
-    return new TempestFeed(WEATHER_TEMPEST_API_KEY, weatherUnits);
-#elif WEATHER_VISUALCROSSING_FEED
-    return new VisualCrossingFeed(WEATHER_VISUALCROSSING_API_KEY, weatherUnits);
-#endif
+    switch (WEATHER_FEED) {
```
flattermann, line 45: I'm not 100% sure, but I think a `switch` might cause some problems here, because it might compile/link the unnecessary feeds as well? Needs testing.

We could switch to `#if WEATHER_FEED == VISUALCROSSING` if necessary (but we need to `#define VISUALCROSSING 0` because enum items can't be used by the preprocessor)

flattermann (changes requested):
I would also comment out the `WEATHER_*` #defines in config.h.template so that the user can enable what he needs.

And move the default values to `config.system.h` (if #ifndef) because we might otherwise get compilation errors.

firmware/src/widgets/weatherwidget/WeatherWidget.cpp
nconner-06: Thanks for your feedback, I'll look at your comments.

Did you look at the weather feed .h file, each has a #ifndef section to define the variables, thus I don't know if we need it in config.system.h as well

firmware/config/config.h.template
nconner-06: the API key for VisualCrossing should be user editable - people can get their own keys (if they want) - like for the other feeds.  It's just we provide a key to get them started..

firmware/src/widgets/weatherwidget/WeatherWidget.cpp
nconner-06: As the header files for all the feeds are included in WeatherWidget.cpp, it looks like all the feeds get compiled and linked - even if I change to a "#if" condition...
image

flattermann: Right. If the defaults are already set in the individual feed .h, we don't need them in config.system.h.

Maybe include the feeds conditionally as well?

firmware/config/config.h.template
nconner-06: moved to WeatherWideget.h

firmware/config/config.h.template
nconner-06: all URLs moved to their respective feed.h files

firmware/config/config.h.template
nconner-06: VisualCrossing API key left in config.h.template because users can have their own, but the default (provided) key is in the visualcrossing feed with an #ifndef condition

firmware/config/config.h.template
nconner-06: all defaults are in their respective feed.h file, all customizable variables are in the config.h.template.
for me, config.systel.h should be for non-widget variables

firmware/config/config.h.template
nconner-06: default is set in WeatherWidget.h, just below the enum which defines the available values

firmware/src/widgets/weatherwidget/WeatherWidget.cpp
nconner-06: I tested both way - switch and ifdef (including ifdef on the fee.h files) and it appears that all feeds are included in the compile / ink.. not sure if I was doing any incorrect,  if you know a way, then I'm open to suggestions.

PR #237, opened 2025-02-19
Features for ClockWidget and WeatherWidget

nconner-06:
As posted on Discord, I've tried generating a pull request for the additional features I have developed
- Changed the API used for TZ to not require an API key
- Added support for multiple TZ - up to 5 in total
- Added a new clock type MORPH
- Added support for a Second Weather location
- Added support for OpenWeatherMap API
- Added support for Lat/Long locations
- Added addition functionality to the web interface to support these new functionalities
- Added minute support to the DIM function
- Changed the Host name of the Orb to "ESP32-Orb"
- And a few other minor changes.

PR #303, opened 2025-03-18
added Morph clock type

nconner-06:
Added Morph clock type to Clock widget

I finally found time to split the Morph clock type out of my big PR (Features for ClockWidget and WeatherWidget #237) into a smaller package


firmware/src/widgets/clockwidget/ClockWidget.h
```diff
@@ -56,27 +56,32 @@
...
-    CUSTOM1 = 3,
-    CUSTOM2 = 4,
-    CUSTOM3 = 5,
-    CUSTOM4 = 6,
-    CUSTOM5 = 7,
-    CUSTOM6 = 8,
-    CUSTOM7 = 9,
-    CUSTOM8 = 10,
-    CUSTOM9 = 11
+    MORPH = 1,
```
ericthelin: I think we should maintain the order of normal then, nixie, then morph. If we keep adding clock types, which I think would be a good thing, we may need to add a more advanced architecture more like widgets where users can choose which to enable

ericthelin (commented):
I haven't gotten a chance to test this, which I would want to do before approving it. But in the mean time please update it and resolve the conflict.

firmware/src/widgets/clockwidget/ClockWidget.h
nconner-06: It doesn't make a lot of difference to me, so switch nixie and morph in the .h file

Issue #304, opened 2025-03-18
5Zone widget and Async HTTP requests


dreed47:
While testing the 5zone widget, startup seems to take a lot longer due to waiting for initial time zone API requests.   Those API requests also seem to be locking up the UI in some cases.  These requests should be done async using TaskManager so startup is not delayed and UI doeesnt get locked up.  Also should include a default Loading message while the API's are returning.

HugoBalder:
Oh boy, the code quality and testing before even thinking to create a Pull Request is getting lower and lower when i once created my PR‘s :( Maybe it will get better in the future.

brett-dot-tech:
Thanks for this @dreed47, my bad for not catching this while merging. I'll work to see if I can fix on my end.

@HugoBalder while I appreciate the sentiment of your comment it's not helpful.
This is an open source project with folks of varying skill levels (of which I am on the lower end of) that I am doing my best to accommodate and make sure all willing to contribute are able to do so, and I'm still learning!

Suggestions, feedback, changes and help are VERY welcomed(+encouraged!!) and I do my best to accommodate any and all of them, however empty criticism is not helpful.

nconner-06:
Hi @dreed47 
Thanks for the feedback, I'll have a look at moving to TaskManager for the API loading.
One point to consider is what is better to have a longer load time and showing correct data or loading quickly and not showing correct because the TZs have not been loaded?
I see this sometimes on the stock widget, if you start your Orb and quickly scroll to the stockwidget you get the "loading data" message flashes across the screens very quickly..
At present there is the standard loading message as per the other widgets, what message are you referring too?

dreed47:
I didn't realize there was already a loading messages. I just didn't see them.  I get your point about having data preloaded but as a general rule any long running operation like an http request should be done async otherwise it completely blocks the main loop and everything appears locked up until it finishes. That's compounded when we have to do multiple http requests one for each screen

dreed47:
If we really wanted to hold up the startup until all data is ready we should use a semaphore

flattermann:
I would prefer the startup to **not** be blocked, but to load all data in a background thread.

Especially if we add more widgets, the time-to-fully-load-all-widgets will go up, but IMO it's better if the user can at least see the clock while the data is still loading.

nconner-06:
I'm re-writing the startup to use the task manager, so as not to block the Orb

Also, building in the new timer function for the draw frequency

nconner-06:
[...]

PR #305, opened 2025-03-20
update to cover : 5Zone widget and Async HTTP requests #304

nconner-06:
5Zone widget and Async HTTP requests #304

### DIFF
--- a/firmware/src/widgets/5zonewidget/5ZoneWidget.cpp
+++ b/firmware/src/widgets/5zonewidget/5ZoneWidget.cpp
@@ -15,22 +15,22 @@ FiveZoneWidget::FiveZoneWidget(ScreenManager &manager, ConfigManager &config) : 
 
     for (int i = 0; i < MAX_ZONES; i++) {
         const char *zoneName = strdup((String("5zoZoneName") + String(i)).c_str());
-        const char *zoneDesc = strdup((i18nStr(t_5zoneDesc) + " " + String(i) + ": ").c_str());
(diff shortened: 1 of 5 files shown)


firmware/src/widgets/5zonewidget/5ZoneWidget.cpp
```diff
@@ -69,6 +69,15 @@ void FiveZoneWidget::processResponse(TimeZone &timeZone, int httpCode, const Str
...
                     timeZone.nextTimeZoneUpdate = makeTime(m_temp_t) + random(5 * 60); // Randomize update by 5 minutes to avoid flooding the API;
+
+                    int lv_idx = 0;
+                    do {
+                        if (m_timeZones[lv_idx].tzInfo == timeZone.tzInfo) {
+                            m_timeZones[lv_idx].timeZoneOffset = timeZone.timeZoneOffset;
+                            m_timeZones[lv_idx].nextTimeZoneUpdate = timeZone.nextTimeZoneUpdate;
+                        }
+                        lv_idx++;
+                    } while (lv_idx < MAX_ZONES);
```
lyricnz, line 80: Could be a for loop?  If there can only be one match, it should `break` also.

nconner-06: yes, it could be a 'for' loop..  not sure if there's that much difference in terms of cpu and memory.. 

No, it should not break...  There could be multiple zones with the same tzInfo, so have to check all of them..
For example, if zones 1, 2 and 5 are the same...

dreed47:
Could we add a persistent config variables for the offsets and on startup check to see if you already have the offset and only go to the API if you need to?

PR #308, opened 2025-03-27
Increase widget limit and do not add disabled widgets

flattermann:
The maximum number of widgets has been raised from 5 to 10 to allow greater flexibility. A check was also added to log and prevent adding disabled widgets, improving robustness and debugging capabilities.

### DIFF
--- a/firmware/src/core/widget/WidgetSet.h
+++ b/firmware/src/core/widget/WidgetSet.h
@@ -5,7 +5,9 @@
 #include "Utils.h"
 #include "Widget.h"
 
-#define MAX_WIDGETS 10
+#ifndef MAX_WIDGETS
+    #define MAX_WIDGETS 5
+#endif
 
 class WidgetSet {
 public:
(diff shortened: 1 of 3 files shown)


firmware/src/core/widget/WidgetSet.h
```diff
@@ -5,7 +5,7 @@
 #include "Utils.h"
 #include "Widget.h"
 
-#define MAX_WIDGETS 5
+#define MAX_WIDGETS 10
```
dreed47: conceptually I agree with making this larger than 5 but we clearly don't have the memory to run 10.  I find that running 7 causes all kinds of memory related errors.  Maybe we should make this 6 or 7 instead?

flattermann: I think it depends a lot on the actual widget. A fixed value will probably never be correct.

Also, at the moment, we fully construct a widget even if it's not enabled. 
It's just not added to WidgetSet and will not receive .setup()/.update()/.draw() calls.
Also all GUI parameter are loaded and each one consumes ~100 bytes or more of RAM.

We should probably optimize that in the future.

I'll move this setting to config.system.h and keep it at 7 for now.
It can be overriden in config.h if necessary.

Issue #313, opened 2025-03-30
Memory allocation errors


dreed47:
PR #268 changed the Async stack size from 6000 to 12000.  This is causing the app to run out of memory in certain situations.

```bash
00:00:30.283 I: 🔵 Starting HTTP request for:.....
[ 30356][E][ssl_client.cpp:37] _handle_error(): [start_ssl_client():264]: (-32512) SSL - Memory allocation failed
[ 30367][E][WiFiClientSecure.cpp:144] connect(): start_ssl_client: -32512
00:00:30.377 E: 🔴 HTTP request failed, error code: -1
```

This should be changed back to 6000

dreed47:
fix merged

PR #268, opened 2025-03-06
Pagination of securities

dgutzmann:
# Description
Allow more than 5 securities via pagination.

## Changes
* Increase task manager params to support more requests
  * There could me more optimization here to make it simpler and based on the total max requests needed
* Moved the initialization bool to be a part of the stock data model instead of the stock widget
* Fixed a check for m_stocks check to ensure that we don't try access past end of array

PR #314, opened 2025-03-30
Quick fix on Async Task stack size

dreed47:
Per issue #313 PR #268 changed the Async stack size from 6000 to 12000. This is causing the app to run out of memory in certain situations.  This change moves the stack size back to 6000


dgutzmann:
Apologies, I was having different issues at one point and I made that value double.

PR #319, opened 2025-04-05
New WidgetRegistry - Only include non disabled widgets in the compiled binary

dreed47:
**Current Issue:**  
As we add more widgets, our compiled binary grows because every widget is included in the compilation, even disabled ones.  

**Solution:**  
This change moves widget registration out of `main.cpp` and into a dedicated file (`WidgetRegistry.cpp`). Now:  
- Only enabled widgets are included in the build.  
- `main.cpp` is cleaner—it just needs to call `registerWidgets(widgetSet, sm, config)`.  

This change also adds a couple build flags to the platformio.ini to help reduce the final binary size by eliminating unused code and data.  And also switches the platformio.ini to use extends.

**How to Add a New Widget:**  
1. Add your widget code to the repo.  
2. Register it in `WidgetRegistry.cpp` and `config.system.h`.  

This makes the build more efficient and keeps the codebase better organized.

### DIFF
--- a/firmware/src/core/widget/WidgetRegistry.cpp
+++ b/firmware/src/core/widget/WidgetRegistry.cpp
@@ -1,0 +1,68 @@
+#include "WidgetRegistry.h"
+
+// Include widget headers
+#include "clockwidget/ClockWidget.h"
+#if INCLUDE_WEATHER != WIDGET_DISABLED
+    #include "weatherwidget/WeatherWidget.h"
+#endif
+#if INCLUDE_STOCK != WIDGET_DISABLED
+    #include "stockwidget/StockWidget.h"
+#endif
+#if INCLUDE_PARQET != WIDGET_DISABLED
+    #include "parqetwidget/ParqetWidget.h"
+#endif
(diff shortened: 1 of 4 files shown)


flattermann:
Actually, disabled widgets should not increase the compiled size, because they will not be linked.
This will work, even if you `#include` them, i.e. you do not need to `#if` the `#include` (just the WidgetSet->add).

That being said, I'm fine with moving `addWidgets()` to a separate file as this will make the `main.cpp` cleaner.

dreed47:
> Actually, disabled widgets should not increase the compiled size, because they will not be linked. This will work, even if you `#include` them, i.e. you do not need to `#if` the `#include` (just the WidgetSet->add).
> 
> That being said, I'm fine with moving `addWidgets()` to a separate file as this will make the `main.cpp` cleaner.

I thought I saw a difference in the binary size between current and when I wrap the includes with '#if'.  I don't have exact stats but i can check again later. But also, when i moved the widgets includes to '#if' I get compile errors because some libraries like TaskFactory and GlobalTime were no longer linked in so in dev they are getting linked in through their references in the widgets somehow.  This change should make things more explicit.

PR #321, opened 2025-04-20
stock state machine

dgutzmann:
Switching to a state machine for the stock ticker status.

This allows to consolidate the drawing of screens and allows to display more information such as a bad security symbol


ericthelin (commented):
I am not the stock ticker expert but this looks good to me.

PR #326, opened 2025-07-19
Implement automatic dimming capability

### LINKED ISSUE #325: Automatic Dimming
Support automatic dimming of the displays in response to changes in room lighting. Room lighting level can be sensed with a simple hardware modification that entails adding a photoresistor and a fixed resistor to the break out area on the info orbs circuit board. Software changes will allow periodic checking of the room lighting and adjusting the displays accordingly.

shemers42:
Add support for automatic dimming to adjust for changing room brightness. Requires a minor hardware modification described in [ReadMe.txt](https://github.com/user-attachments/files/21323920/ReadMe.txt) .

 This software change can be safely merged since it is disabled by default, and will have no effect unless `AUTO_DIMMING_ENABLED` is set to `true` in config.h

Closes #325

https://github.com/user-attachments/assets/0e3514b4-0801-445c-8f01-bd8083e6f8c7

### DIFF
--- a/firmware/config/config.h.template
+++ b/firmware/config/config.h.template
@@ -29,6 +29,12 @@
 // #define DIM_END_HOUR 7     // Undim the screens at this time (24h format)
 // #define DIM_BRIGHTNESS 128 // Dim brightness (0-255)
 
+// AUTO DIMMING (requires hardware mod -- SEE config.system.h)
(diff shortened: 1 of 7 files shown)


ericthelin:
This is an interesting idea. The PR is well done. But I question if we want to add code that requires a hardware modification. @brett-dot-tech I think this one is clearly in your court to give direction on.

PR #327, opened 2026-02-06
Nhl

brtcruiser1234:
This is based on the Baseball Widget but for NHL


ericthelin:
This has branch/merge issues and is showing many unrelated commits as part of the PR. I attempted to change the target branch to `dev` to see if it would help and I think it helped, but not enough. Please rework the PR to only include your actual changes vs the `dev` branch and we can review it for merging

PR #331, opened 2026-07-07
Feature/calendar widget

SimonGeiger:
implemented and tested calendar widget


README.md
SimonGeiger, line 1: no need for name

PR #333, opened 2026-07-14
wifi power level escalation to reduce brownouts

ericthelin:
I have been running my orbs on a power adapter on the edge of what is required because the ESP32 doesn't request higher current from modern usb-c chargers which means that sometimes starting wifi on boot causes a brownout event rebooting the device and once it starts that loop it will reboot infinitely until unplugged. This is an attempt to address that by observing the signal level for the configured wifi and matching the transmit power to what is needed.


nconner-06:
Hi
Small point, why are you using "Serial.printf" and not "Log.infoln"?

PR #335, opened 2026-07-15
Fix clock colon blink regularity and AM/PM redraw heap churn

Bullpuph123:
Split from #330 (part 2 of 3), rebased onto dev as requested.

## Summary
- **Smooth, drift-free colon blink (NORMAL clock)**: the colon was driven by seconds parity from GlobalTime, so any skipped/duplicated second made the blink irregular. It's now driven by a millis()-based 500ms blinker with parity-correct catch-up (if draw is delayed by multiple periods, visibility only toggles on an odd number of elapsed periods). NIXIE/custom clocks keep the image-based 1s colon.
- **AM/PM redrawn only on change**: `displayAmPm()` was called every second, and each call reloads TTF fonts (DSEG7↔DSEG14) plus per-loop String allocations — over days of uptime this fragments the heap and eventually corrupts glyph rendering (stray "PM" artifacts observed after long uptime). It's now redrawn only when the value changes or on force, with the old value cleared first.
- **Bounds/ownership guards**: `displayDigit()` validates the display index; `displaySeconds()` only draws on the status orb and validates the seconds range — defense against a corrupted index painting tick marks on the wrong panel.
- `CLOCK_UPDATE_DELAY`/`CLOCK_DRAW_DELAY` defaults changed from `OneSecond` to `OneHundredMilliseconds` so the 500ms colon blinker actually gets a chance to run; all per-second work is still internally gated on second changes, so this doesn't add redraw load.

## Test plan
- [x] Builds with `pio run -e infoorbs-v0_3`
- [x] Equivalent change soak-tested on hardware for several days on the main-based branch from #330 (colon blinks evenly, no AM/PM artifacts)
- [x] Verify NIXIE/custom clock colon still alternates each second

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/firmware/src/widgets/clockwidget/ClockWidget.cpp
+++ b/firmware/src/widgets/clockwidget/ClockWidget.cpp
@@ -65,6 +65,10 @@ void ClockWidget::setup() {
     m_lastDisplay2Digit = "";
     m_lastDisplay4Digit = "";
     m_lastDisplay5Digit = "";
+    m_lastAmPm = "";
+    m_lastSecondSingle = -1;
+    m_colonVisible = true;
+    m_colonBlinkPrev = millis();
 }
 
 void ClockWidget::draw(bool force) {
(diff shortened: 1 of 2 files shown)


firmware/src/widgets/clockwidget/ClockWidget.h
```diff
@@ -89,8 +89,13 @@ class ClockWidget : public Widget {
     String getName() override;
 
 private:
+    // The middle/status orb exclusively owns the colon, the seconds tick
+    // and the AM/PM indicator. Digit orbs (0, 1, 3, 4) must never receive them.
+    static constexpr int SCREEN_STATUS = 2;
```
ericthelin: I don't love the variable name SCREEN_STATUS because "status" is not really what is being displayed

ericthelin (commented):
I would like to see the constant renamed then I will approve this

firmware/src/widgets/clockwidget/ClockWidget.h
Bullpuph123: Good point — renamed to SCREEN_MIDDLE in 2b44a84, matching the existing BUTTON_MIDDLE naming convention.

firmware/src/widgets/clockwidget/ClockWidget.cpp
```diff
@@ -88,35 +92,73 @@ void ClockWidget::draw(bool force) {
...
+            // network call) can never invert the blink phase.
+            uint32_t periods = elapsed / m_colonBlinkInterval;
+            m_colonBlinkPrev += m_colonBlinkInterval * periods;
+            if (periods % 2 == 1) {
+                m_colonVisible = !m_colonVisible;
+            }
+            displayColon();
+        } else if (force) {
+            displayColon();
+        }
```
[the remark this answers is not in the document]
Bullpuph123: Fixed in 2b44a84: forced redraws now restart the blink phase with the colon visible (m_colonVisible = true, m_colonBlinkPrev = now) instead of running the catch-up path, so the colon can no longer reappear in the "off" state after a screen clear or widget switch.

